PR #62, opened 2017-06-15
💄 - Migrate to jest

### LINKED ISSUE #52: Test Migration
Migrate tests to jest or similar. The current test setup is really awkward to work with.

arb:
Closes #52

Migrated all tests to use jest instead of mocha. Removed all browser testing
code. Updated tests so that each spec test that needs a database gets it's
own database so one test doesn't impact others in a different spec.

Migrated fixtures to be nested under tests.

Removed shell scripts as they were no longer being used.

Cleanup for package.json.

Prettier run across the test files.


test/changePwd.spec.js
```diff
@@ -1,81 +1,52 @@
...
+    const password = generateRandomString();
+    conn.createUser(
+      { username, password, superuser: true },
+      (data, response) => {
+        // It should be 201 (CREATED)
+        expect(response.statusCode).toEqual(201);
+
+        conn.changePwd(
+          { user: username, newPwd: password },
+          (data1, response1) => {
```
snowell: We're changing the password to the old password here. Should generate a new one

test/copyDB.spec.js
```diff
@@ -1,87 +1,60 @@
...
+} = require('./setup-database');
+
+describe('copyDB()', () => {
+  const sourceDatabase = generateDatabaseName();
+  const destinationDatabase = generateDatabaseName();
+  var conn;
+
+  beforeAll(seedDatabase(sourceDatabase));
+  afterAll(dropDatabase(sourceDatabase));
+  afterAll(dropDatabase(destinationDatabase));
```
snowell: Should these be combined into one afterAll call?

test/copyDB.spec.js
```diff
@@ -1,87 +1,60 @@
...
+    conn = new Stardog.Connection();
+    conn.setEndpoint('http://localhost:5820/');
+    conn.setCredentials('admin', 'admin');
+  });
+
+  it('should not copy an online DB', done => {
+    conn.onlineDB(
+      { dbsource: sourceDatabase, strategy: 'WAIT', timeout: 3 },
+      () => {
+        conn.copyDB(
```
snowell: No assertions on the return code?

test/copyDB.spec.js
```diff
@@ -1,87 +1,60 @@
...
+      () => {
+        // Once the DB is offline, copy it.
+        conn.copyDB(
+          { dbsource: sourceDatabase, dbtarget: destinationDatabase },
+          () => {
+            conn.listDBs(data => {
+              expect(data.databases).toEqual(
+                expect.arrayContaining([sourceDatabase, destinationDatabase])
+              );
+              done();
```
snowell: No assertions on the response code?

test/createDB.spec.js
```diff
@@ -1,106 +1,34 @@
...
+    };
+
+    conn.createDB(options, (data, response) => {
+      expect(response.statusCode).toEqual(409);
+
+      done();
     });
-}));
\ No newline at end of file
+  });
```
snowell: Are we just punting on the "should create a new empty DB, returning 201" test?

test/dropDB.spec.js
```diff
@@ -1,73 +1,20 @@
...
-        });
+  it('should not drop an non-existent DB', done => {
+    conn.dropDB({ database: 'nodeDBDrop' }, (data, response) => {
+      expect(response.statusCode).toEqual(404);
 
+      expect(data).toContain('does not exist');
+      done();
     });
-}));
+  });
```
snowell: What about "should drop a just created database?"

test/getDBOptions.spec.js
```diff
@@ -1,81 +1,58 @@
...
+  beforeAll(seedDatabase(database));
+  afterAll(dropDatabase(database));
+
+  beforeEach(() => {
+    conn = new Stardog.Connection();
+    conn.setEndpoint('http://localhost:5820/');
+    conn.setCredentials('admin', 'admin');
+  });
+
+  it('should get NOT_FOUND status code trying to get the options of a non-existent DB.', done => {
```
snowell: It doesn't seem like this is what is actually being tested. It looks like this test is expecting a 404 for onlineDB, and then a 500 for getDBOptions

test/getDBSize.spec.js
```diff
@@ -1,56 +1,29 @@
...
+  beforeEach(() => {
+    conn = new Stardog.Connection();
+    conn.setEndpoint('http://localhost:5820/');
+    conn.setCredentials('admin', 'admin');
+  });
+
+  it('A response with the size of the DB should not be empty', done => {
+    conn.getDBSize({ database }, response => {
+      var sizeNum = parseInt(response, 10);
+      expect(sizeNum).toBe(48);
```
snowell: Why change from a test of X>0? This way if the db changes in any way this can bomb for no reason

test/isSuperUser.spec.js
```diff
@@ -1,60 +1,32 @@
...
+  });
 
+  it("should return the value with the user's superuser flag", done => {
+    conn.isSuperUser({ user: 'admin' }, (data, response) => {
+      expect(response.statusCode).toEqual(200);
+      expect(data.superuser).toEqual(true);
+      done();
     });
-}));
+  });
```
snowell: Very pedantic here, but should we also test the case where the user is NOT a superuser?

test/isUserEnabled.spec.js
```diff
@@ -1,59 +1,26 @@
...
+  });
 
+  it("should return the value with the user's enabled flag", done => {
+    conn.isUserEnabled({ user: 'admin' }, (data, response) => {
+      expect(response.statusCode).toEqual(200);
+      expect(data.enabled).toEqual(true);
+      done();
     });
-}));
+  });
```
snowell: Again, should we test for the case when this should return false?

test/listRoleUsers.spec.js
```diff
@@ -1,53 +1,34 @@
...
+  beforeAll(seedDatabase(database));
+  afterAll(dropDatabase(database));
+
+  beforeEach(() => {
+    conn = new Stardog.Connection();
+    conn.setEndpoint('http://localhost:5820/');
+    conn.setCredentials('admin', 'admin');
+  });
+
+  it("should return a list of users assigned to the 'reader' role in the system.", done => {
```
snowell: Unimportant, but we aren't actually testing the 'reader' role here

test/query.spec.js
```diff
@@ -1,429 +1,291 @@
...
+
+  it('A query result should not be empty', done => {
+    conn.onlineDB({ database, strategy: 'NO_WAIT' }, () => {
+      conn.query(
+        {
+          database,
+          query: 'select distinct ?s where { ?s ?p ?o }',
+        },
+        data => {
+          expect(data.results.bindings).toHaveLength(10);
```
snowell: Magic numbers aren't great, not that I have a better alternative here

test/query.spec.js
```diff
@@ -1,429 +1,291 @@
...
+              '  OPTIONAL { ?s rdf:type ?sType .\n' +
+              '            OPTIONAL { ?sType rdfs:label ?sTypeLabel . }\n' +
+              '           }\n' +
+              '  OPTIONAL { ?o rdf:type ?oType .\n' +
+              '            OPTIONAL { ?oType rdfs:label ?oTypeLabel . }\n' +
+              '           }' +
+              '}',
+        },
+        data => {
+          expect(data.results.bindings).toEqual(expect.anything());
```
snowell: Would this return true if data.results.bindings equalled {}? Because that's almost certainly not desired

test/query.spec.js
```diff
@@ -1,429 +1,291 @@
...
+    conn.onlineDB({ database, strategy: 'NO_WAIT' }, () => {
+      conn.query(
+        {
+          database,
+          query: 'select * where { ?s ?p ?o }',
+          limit: 10,
+          offset: 0,
+        },
+        data => {
+          expect(data.results.bindings).toHaveLength(10);
```
snowell: In theory, you could test for a length <= 10 and be true to the test

test/query.spec.js
```diff
@@ -1,429 +1,291 @@
...
+      conn.query(
+        {
+          database,
+          query: 'select * where { <Article1> ?p ?o }',
+          baseURI: 'http://localhost/publications/articles/Journal1/1940/',
+          limit: 10,
+          offset: 0,
+        },
+        data => {
+          expect(data.results.bindings).toHaveLength(10);
```
snowell: This is not testing anything having to do with baseURI

test/queryList.spec.js
```diff
@@ -1,49 +1,24 @@
...
-                expect(data.queries instanceof Array).to.be(true);
-                expect(data.queries.length).to.be(0);
-                done();
-            });
-        });
+  it('should start a query and list it with the listQueries call.', done => {
+    conn.queryList((data, response) => {
+      expect(response.statusCode).toEqual(200);
+      expect(data.queries).toHaveLength(0);
+      done();
```
snowell: I get that this is styled after the original test...but it's not doing anything at all like it says it is.

test/setUserRoles.spec.js
```diff
@@ -1,67 +1,48 @@
...
+        password,
+        superuser: true,
+      },
+      () => {
+        conn.setUserRoles(
+          { user: username, roles: ['reader'] },
+          (data, response) => {
+            expect(response.statusCode).toEqual(200);
+            done();
+          }
```
snowell: Do we not want to check that the role is, in fact, assigned?

test/assignPermissionToRole.spec.js
```diff
@@ -1,77 +1,59 @@
...
+  var conn;
+
+  beforeAll(seedDatabase(database));
+  afterAll(dropDatabase(database));
+
+  beforeEach(() => {
+    conn = new Stardog.Connection();
+    conn.setEndpoint('http://localhost:5820/');
+    conn.setCredentials('admin', 'admin');
+  });
```
arb: That's how the original was. I removing the `null` part because it wasn't needed as the reference was being replace after every test anyway.

It's in `beforeEach` because we want to make sure no change in any tests carry across to other tests. It's a good practice. Either do it this way, or create a factory function and create a new connection in each test and drop the `beforeEach` entirely.

test/assignPermissionToRole.spec.js
```diff
@@ -1,77 +1,59 @@
...
+    conn.createRole({ rolename: aNewRole }, (data, response1) => {
+      expect(response1.statusCode).toEqual(201);
+
+      conn.assignPermissionToRole(
+        { role: aNewRole, permissionObj: aNewPermission },
+        (data, response2) => {
+          expect(response2.statusCode).toEqual(201);
+          done();
+        }
+      );
```
arb: I dropped all that cleanup code because we destroy the entire database after each spec is complete. Makes the tests much more manageable.

test/copyDB.spec.js
arb: Wouldn't really matter and it would make the code more verbose to do it :-D

test/createDB.spec.js
arb: For now yes because we do those things over and over again in each test. I test to make sure the status code is 201.

test/dropDB.spec.js
arb: Same as above.

test/getDBSize.spec.js
arb: It's a tighter test that way. And regarding your second point, that's why I don't really want to be actually hitting a stardog server with any of these tests 😄

test/query.spec.js
arb: How would I test that? Not sure what `baseURI` is or how it affects the query.

test/queryList.spec.js
arb: What is your suggestion for addressing that?

snowell: Have it kick off a long-enough-running query such that you can assert data.queries to have length of 1. This test currently is simply testing that queryList says there are 0 queries running when there are 0 queries running

test/query.spec.js
```diff
@@ -1,429 +1,271 @@
...
+        offset: 0,
+      },
+      data => {
+        expect(data.results.bindings).toHaveLength(10);
+        // console.log(require('util').inspect(data.results, true, null, true))
+        done();
+      }
+    );
+  });
+
```
snowell: Can't reply to the original message for some reason. But regarding baseURI:

https://www.w3.org/TR/2013/REC-sparql11-query-20130321/#relIRIs explains what this is. It's essentially another way to shorten IRIs without using prefixes.

In the case of your data, the query `select * {?s a <http://localhost/vocabulary/bench/Article>}` returns 3 results. By including baseURI=http://localhost/vocabulary/bench/, you could shorten the query to `select * {?s a <Article>` and get the same results. This test should test that.

snowell: It's basically another way of defining the `:` prefix and doing `select * {?s a :Article}`

test/queryList.spec.js
arb: We decided to leave this alone as we couldn't write a query that took long enough for the results to ever be more than 0.

test/query.spec.js
arb: Ok done!

PR #63, opened 2017-06-20
Rollup integration

### LINKED ISSUE #49: Rollup integration
Use webpack to manage bundling, minification, and library definitions.

### LINKED ISSUE #58: Migrate to ESLint Airbnb style

tysoeh:
adds rollup, eslint, babel, and uglify. closes #49 and closes #58

creates a build script which generates rollup bundles for umd and esm formats, copies the license and a trimmed package.json


scripts/build.js
```diff
@@ -0,0 +1,77 @@
...
+        .then(build => build.write({
+          dest: bundle.dest,
+          format: bundle.format,
+          moduleName: bundle.moduleName || 'main',
+          sourceMap: !bundle.minify || bundle.sourceMap || false,
+        }))
+        .then(() => console.log(`${bundle.dest} written successfully.`));
+    });
+  })
+  .then(() => {
```
arb: Cute idea, but atypical and unneeded complexity.

scripts/build.js
```diff
@@ -0,0 +1,77 @@
...
+const fs = require('fs');
+const del = require('del');
+const rollup = require('rollup');
+const babel = require('rollup-plugin-babel');
+const eslint = require('rollup-plugin-eslint');
+const uglify = require('rollup-plugin-uglify');
+const pkg = require('../package.json');
+
+const bundles = [
+  {
```
arb: Can you explain why we are publishing 3 different formats?

scripts/build.js
```diff
@@ -0,0 +1,77 @@
...
+
+const fs = require('fs');
+const del = require('del');
+const rollup = require('rollup');
+const babel = require('rollup-plugin-babel');
+const eslint = require('rollup-plugin-eslint');
+const uglify = require('rollup-plugin-uglify');
+const pkg = require('../package.json');
+
+const bundles = [
```
arb: This is supposed to be usable both on the client and in node... I don't think we need `babel` for the node version.

scripts/build.js
```diff
@@ -0,0 +1,77 @@
...
+        exclude: 'node_modules/**',
+        presets: ['es2015-rollup'],
+      }),
+      uglify(),
+    ],
+    moduleName: 'stardog',
+    minify: true,
+  },
+  {
+    format: 'umd',
```
arb: Why two UMDs?

scripts/build.js
```diff
@@ -0,0 +1,77 @@
...
+          moduleName: bundle.moduleName || 'main',
+          sourceMap: !bundle.minify || bundle.sourceMap || false,
+        }))
+        .then(() => console.log(`${bundle.dest} written successfully.`));
+    });
+  })
+  .then(() => {
+    delete pkg.devDependencies;
+    delete pkg.scripts;
+    pkg.main = 'stardog.min.js';
```
arb: Since these filenames don't change, you could just update the real `package.json`. Also, look into the `browser` field as well.

scripts/build.js
```diff
@@ -0,0 +1,77 @@
...
+  {
+    format: 'umd',
+    dest: 'build/stardog.min.js',
+    plugins: [
+      eslint(),
+      babel({
+        exclude: 'node_modules/**',
+        presets: ['es2015-rollup'],
+      }),
+      uglify(),
```
arb: We'll also want to publish a non-minifield version. Most people (us included) just minify everything in one shot so our users will want a non-minified version as well.

scripts/build.js
tysoeh: which idea? making a trimmed copy of the manifest, or adding it to the promise chain?

scripts/build.js
tysoeh: one is minified, one is not

scripts/build.js
tysoeh: umd minified/!minified, es modules

scripts/build.js
tysoeh: i think it makes more sense to remove `main` and `module` from the real package.json and declare them here. they shouldn't be used from the root manifest, and if they were the paths would have to be different than the `dist/` manifest

scripts/build.js
tysoeh: we'll at least need to use babel to transpile from esm imports/exports to commonjs require/exports

scripts/build.js
arb: Trimmed copy of the manifest.

scripts/build.js
arb: I wasn't planning on using anything other than common JS though.

arb:
This also closes #59

PR #68, opened 2017-06-30
Cleanup part 1

arb:
Stopping here to get feedback from others. Currently, the following tests all

## Commentary

For now, I'm keeping the old index.js file around for reference as it's easier to look at that then to try to figure out HTTP calls to the server. 

One thing I'd like to comment on is `params`. All of the methods that hit the HTTP server have `params` as a final argument. The intent there is to (eventually) allow users to pass any additional query parameters along. As I don't know what they could possibly be, I haven't really explored what they could be, but that's why it's there but not currently used.

Regarding my `FormData` TODO comment... all of these test are currently running under the node context and are working. In jest you can pass the `--browser` flag which will cause the `require` system to honor the "browser" field in package.json when it tries to load things in. This will give you the browser version of modules that have that set up. Going down that execution path, we end up using jsdom's version of an `XMLHTTPRequest`. Per this comment here https://github.com/tmpvar/jsdom/blob/7d950c448fda038c8c385e5fba07fd85bd19326e/lib/jsdom/living/xmlhttprequest.js#L578-L579, I feel like multipart form data with proper content-type and boundary headers do not work with the version of jsdom that jest is currently using. This matters because... when the fetch pony fill I'm using fires up, going down the browser path, it implements the `fetch` API via promises and `XMLHTTPRequests`. That means it's using the incomplete jsdom version and NOT a full browser implementation. I am confident that running this code, properly bundled, in a browser would function. The only API that's even remotely complex is creating a DB, everything else is simple. The issue is that EVERY tests uses the create database API so if that doesn't work, none of the tests work.

The *TL;DR* is, I'm going to continue writing and running tests against Node for now with the understanding that we're going to have to revisit this. To this point, that weird multi-part form data thing is the ONLY spot where we have issue between node and browser.

## Tests

Currently, these test pass:

`node node_modules/.bin/jest --verbose -i test/changePwd.spec.js test/Connection.spec.js test/copyDB.spec.js test/getDB.spec.js test/getDBOptions.spec.js test/setDBOptions.spec.js test/getProperty.spec.js test/isUserEnabled.spec.js test/listDBs.spec.js test/optimizeDB.spec.js test/query.spec.js test/queryInfo.spec.js test/queryKill.spec.js test/queryList.spec.js`

## Update (2017-07-07)

All tests currently pass. Bulk of API has been completely rewritten. The function contracts aren't in stone yet, but pretty stable at this point. If I deleted a test file, it's because it's contents were migrated to another file. There is certainly more cleanup work to be done, but I want to get this in so the work can be split up across files and people.


lib/db.js
```diff
@@ -0,0 +1,150 @@
+const { fetch } = require('fetch-ponyfill')();
+const FormData = require('form-data');
+const qs = require('querystring');
+const flat = require('flat');
+
+const DB_OPTIONS = require('./dbopts');
+const { httpMessage, httpBody } = require('./response-transforms');
+
+const dbOptions = (conn, options, body) => {
```
tysoeh: What do you think of destructuring `options` here and elsewhere, e.g. `(conn, { method, headers }, body) => {`?

lib/index2.js
```diff
@@ -0,0 +1,11 @@
+const pkg = require('../package.json');
```
tysoeh: What's your thinking behind filtering exports through this file?

lib/db.js
arb: I don't really have a strong preference either way.

tysoeh: My argument for destructuring is that it self-documents which options are needed. Also would make things a little cleaner to be able to use shorthand within the functions.

PR #72, opened 2017-07-07
CircleCI integration tests

tysoeh:
Adds a CircleCI build that pulls an image of Stardog 5 and burns a license and test data onto the image, then deploys the image on port 5820 and runs our jest tests against it.


arb:
This looks amazing to me. Get those tests passing and get travis out of here and we'll be golden.

Issue #75, opened 2017-07-07
Missing Calls


arb:
https://github.com/stardog-union/stardog.js/blob/master/js/stardog.js#L819-L928

Seems to be missing. Not sure if they're still needed or what the status is.

PR #83, opened 2017-07-13
Port old ICV methods

tysoeh:
Rewrites the (known-to-be-working) ICV methods, adds tests, relates to #75

### DIFF
--- a/lib/icv.js
+++ b/lib/icv.js
@@ -1,0 +1,61 @@
+const { fetch } = require('fetch-ponyfill')();
+const qs = require('querystring');
+
+const { httpBody, httpMessage } = require('./response-transforms');
+
+const get = (conn, database, options, params = {}) => {
+  const headers = conn.headers();
+  headers.append('Accept', 'text/plain');
+
+  return fetch(conn.uri(database, 'icv'), {
+    headers,
+  }).then(httpBody);
+};
+
+const set = (conn, database, icvAxioms, options, params = {}) => {
+  const headers = conn.headers();
+  headers.append('Content-Type', options.contentType || 'text/plain');
+
+  return fetch(conn.uri(database, 'icv', 'add'), {
+    method: 'POST',
+    body: icvAxioms,
+    headers,
+  }).then(httpMessage);
+};
+
+const clear = (conn, database, options, params = {}) => {
+  const headers = conn.headers();
+  headers.append('Content-Type', 'application/json');
+
+  return fetch(conn.uri(database, 'icv', 'clear'), {
+    method: 'POST',
+    headers,
+  }).then(httpMessage);
+};
+
+const convert = (conn, database, icvAxiom, options, params = {}) => {
+  const headers = conn.headers();
+  headers.append('Content-Type', options.contentType || 'text/plain');
+
+  const queryParams = {};
+  if (params.graphUri) {
+    queryParams['graph-uri'] = params.graphUri;
+  }
+  const query = qs.stringify(queryParams);
+  const uri = `${conn.uri(database, 'icv', 'convert')}${query.length > 0
+    ? `?${query}`
+    : ''}`;
+
+  return fetch(uri, {
+    method: 'POST',
+    body: icvAxiom,
+    headers,
+  }).then(httpBody);
+};
+
+module.exports = {
+  set,
+  get,
+  clear,
+  convert,
+};
--- a/test/clearICV.spec.js
+++ b/test/clearICV.spec.js
@@ -1,32 +1,37 @@
 /* eslint-env jest */
 
+const fs = require('fs');
+const path = require('path');
 const { icv } = require('../lib');
 const {
   seedDatabase,
   dropDatabase,
   generateDatabaseName,
   ConnectionFactory,
-  icvAxioms,
 } = require('./setup-database');
 
-describe('get integrity constraints', () => {
+const icvAxioms = fs.readFileSync(
+  path.resolve(`${__dirname}/fixtures/issues/constraints.ttl`),
+  'utf8'
+);
+
+describe('icv', () => {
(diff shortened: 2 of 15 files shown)


lib/icv.js
```diff
@@ -0,0 +1,61 @@
...
+
+  return fetch(conn.uri(database, 'icv', 'clear'), {
+    method: 'POST',
+    headers,
+  }).then(httpMessage);
+};
+
+const convert = (conn, database, icvAxiom, options, params = {}) => {
+  const headers = conn.headers();
+  headers.append('Content-Type', options.contentType || 'text/plain');
```
arb: I've been calling this `mimeType`. (I think...)

test/clearICV.spec.js
```diff
@@ -0,0 +1,40 @@
...
+const { icv } = require('../lib');
+const {
+  seedDatabase,
+  dropDatabase,
+  generateDatabaseName,
+  ConnectionFactory,
+  icvAxioms,
+} = require('./setup-database');
+
+describe('get integrity constraints', () => {
```
arb: I think this top level describe should be `icv`

test/clearICV.spec.js
```diff
@@ -0,0 +1,40 @@
...
+    icv
+      .set(conn, database, icvAxioms, { contentType: 'text/turtle' })
+      .then(res => expect(res.status).toBe(204))
+      .then(() => icv.get(conn, database))
+      .then(res => {
+        expect(res.status).toBe(200);
+        expect(res.result.length).toBeGreaterThan(0);
+        expect(
+          res.result.includes('<http://example.org/issues#Issue>')
+        ).toBeTruthy();
```
arb: To be true.

test/clearICV.spec.js
```diff
@@ -0,0 +1,40 @@
...
+  afterAll(dropDatabase(database));
+
+  it('should clear a set of stored axioms from a database', () => {
+    icv
+      .set(conn, database, icvAxioms, { contentType: 'text/turtle' })
+      .then(res => expect(res.status).toBe(204))
+      .then(() => icv.get(conn, database))
+      .then(res => {
+        expect(res.status).toBe(200);
+        expect(res.result.length).toBeGreaterThan(0);
```
arb: Can this be stricter?

test/convertICV.spec.js
```diff
@@ -0,0 +1,26 @@
...
+  const conn = ConnectionFactory();
+
+  beforeAll(seedICVDatabase(database));
+  afterAll(dropDatabase(database));
+
+  it('should store integrity constraint axioms for a database', () => {
+    icv
+      .convert(conn, database, icvAxioms, { contentType: 'text/turtle' })
+      .then(res => {
+        expect(res.result.startsWith('SELECT')).toBeTruthy();
```
arb: To be true.

test/setup-database.js
```diff
@@ -81,3 +122,9 @@ exports.ConnectionFactory = () =>
     endpoint: 'http://localhost:5820',
     // endpoint: 'http://localhost:61941',
   });
+
+exports.icvAxioms = (() =>
```
arb: Why the extra function closure? just `exports.ivcAcioms = fs.readFileSync...`

test/setup-database.js
```diff
@@ -52,6 +53,46 @@ exports.seedDatabase = database => () => {
     });
 };
 
+exports.seedICVDatabase = database => () => {
```
arb: Do we need this? Can't we just also load this file with the normal database seeding? So every database has everything?

test/setup-database.js
arb: Also, I'd rather this get read in for the tests that need it.

test/clearICV.spec.js
tysoeh: I don't think so, but it can probably be removed. The following line asserts that a the result includes a specific string, which takes care of this assertion and greater specificity

test/getICV.spec.js
```diff
@@ -0,0 +1,38 @@
...
+    icv
+      .set(conn, database, icvAxioms(), { contentType: 'text/turtle' })
+      .then(res => expect(res.status).toBe(204))
+      .then(() => icv.get(conn, database))
+      .then(res => {
+        expect(res.status).toBe(200);
+        expect(res.result.length).toBeGreaterThan(0);
+        expect(
+          res.result.includes('<http://example.org/issues#Issue>')
+        ).toBeTruthy();
```
arb: To be true.

test/setup-database.js
```diff
@@ -44,6 +48,57 @@ exports.seedDatabase = database => () => {
...
+          },
+        ],
+      }
+    )
+    .then(res => {
+      expect(res.status).toBe(201);
+    });
+};
+
+exports.seedICVDatabase = database => () => {
```
arb: Do we need this anymore?

test/setup-database.js
```diff
@@ -81,3 +136,9 @@ exports.ConnectionFactory = () =>
     endpoint: 'http://localhost:5820',
     // endpoint: 'http://localhost:61941',
   });
+
+exports.icvAxioms = () =>
```
arb: Why do you want this to be a function? You can just make it the result of the read operation.

test/getDB.spec.js
```diff
@@ -21,7 +21,7 @@ describe('getDB()', () => {
 
   it('A response of the DB info should not be empty', () =>
     db.get(conn, database).then(res => {
-      expect(res.result).toMatch('@prefix : <http://example.org/vehicles/> .');
+      expect(res.result).toMatchSnapshot();
```
arb: Nah. I haven't been using snapshots.

arb:
We should add an issue to keep track of what we're missing.

PR #84, opened 2017-07-13
Headers

### LINKED ISSUE #82: Expose More Headers
There are error codes hidden in response headers that might be useful. Either expose all the response headers or move just the important ones over.

http://www.stardog.com/docs/#_response_codes

arb:
🎨 - Standardize on "body" for response

Renamed "response.result" to "response.body" to be more consistent and
so you don't have to do "response.result.results" for queries.

✨ - More response values

Closes #82.

Attach more information to the response. The response objects now include
status, statusText, ok, url, and headers. The headers object is a bit of an
anomaly as it's a Map on the client, but just a regular object in Node.
It seems like version 2 of node-fetch, will properly implement the headers
object in spec, but it is still in beta.

### DIFF
--- a/lib/response-transforms.js
+++ b/lib/response-transforms.js
@@ -1,26 +1,26 @@
+const lodashPick = require('lodash/pick');
+
+const FIELDS = ['status', 'statusText', 'headers', 'ok', 'url'];
+
 module.exports = {
-  httpMessage: res => ({
-    status: res.status,
-    statusText: res.statusText,
-  }),
+  httpMessage: res => {
+    const response = lodashPick(res, FIELDS);
+    return response;
+  },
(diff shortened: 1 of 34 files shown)


lib/response-transforms.js
```diff
@@ -1,26 +1,27 @@
...
-        const response = {
-          status: res.status,
-          statusText: res.statusText,
-          result,
-        };
+      return res.text().then(text => {
+        response.body = text;
         if (contentType === 'text/boolean') {
-          response.result = result.toLowerCase() === 'true';
+          response.body = text.toLowerCase() === 'true';
```
tysoeh: this is clever. missed it before.

PR #89, opened 2017-07-17
Adds typescript declaration file.

### LINKED ISSUE #50: Typedefs
Add TypScript definitions. Only the defs though, TypeScript would be overkill here.

tysoeh:
Closes #50

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -1,0 +1,286 @@
+// Type definitions for stardog.js 0.3.1
+// Project: https://github.com/stardog-union/stardog.js
+// Definitions by: Stephen Nowell <http://github.com/snowell>, Ty Soehngen <http://github.com/tysoeh>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.1
+
+import Headers from 'fetch-ponyfill';
+import Guid from '@types/node';
+
+declare global {
+  
+  /** stardog.js: The Stardog JS API*/
+  namespace Stardog {
+
+    type ContentType =
+        'application/x-turtle' |
+        'text/turtle' |
+        'application/rdf+xml' |
+        'text/plain' |
+        'application/x-trig' |
+        'text/x-nquads' |
+        'application/trix';
+    
+    type MIMEType =
+        'text/plain' |
+        'application/json' |
+        'text/boolean';
+
+    interface HTTPOptions {
+        headers?: Headers;
+        method?: string;
+        mimeType?: MIMEType;
+        contentType?: ContentType;
+    }
+
+    interface QueryOptions {
+        query: string;
+        baseURI?: string;
+        limit?: number;
+        offset?: number;
+        reasoning?: boolean;
+    }
+
+    interface HTTPMessage {
+        status: string;
+        statusText: string;
+    }
+
+    interface HTTPBody extends HTTPMessage, result {}
+
+    interface result {
+        result: string;
+    }
+
+    interface Params {
+        [paramName: string]: any;
+    }
+    
+    interface ConnectionOptions {
+        endpoint: string;
+        username: string;
+        password: string;
+    }
+
+    /** Describes the connection to a running Stardog server. */
+    class Connection {
+        constructor(ConnectionOptions);
+
+        config(options: ConnectionOptions): void;
+        headers(): Headers;
+        uri(resource: Array<string>): string;
+    }
+    
+    namespace server {
+        /** Shuts down a Stardog server. */
+        function shutdown(conn: Connection, params?: Params) : Promise<HTTPMessage>;
+    }
+
+    /** Stardog database actions. */
+    namespace db {
+
+        interface DatabaseOptions {
+            name: string;
+            archetypes?: string;
+            connection?: timeout;
+            creator?: string;
+            namespaces?: string[];
+            online?: boolean;
+            time?: creation;
+        }
+        interface timeout {
+            timeout: number;
+        }
+        interface creation {
+            creation: string;
+        }
+        /** Creates a new database. */
+        function create(conn: Connection, database: string, databaseOptions?: DatabaseOptions, options?: Object, params?: Params): Promise<HTTPMessage>;
+        /** Deletes a database. */
+        function drop(conn: Connection, database: string, params?: Params) : Promise<HTTPMessage>;
+        /** Gets information about a database. */
+        function get(conn: Connection, database: string, params?: Params) : Promise<HTTPBody>;
+        /** Sets a database offline. */
+        function offline(conn: Connection, database: string, params?: Params) : Promise<HTTPBody>;
+        /** Sets a database online. */
+        function online(conn: Connection, database: string, params?: Params) : Promise<HTTPBody>;
+        /** Optimizes a database. */
+        function optimize(conn: Connection, database: string, params?: Params) : Promise<HTTPMessage>;
+        /** Makes a copy of a database. */
+        function copy(conn: Connection, database: string, destination: string, params?: Params) : Promise<HTTPBody>;
+        /** Gets a list of all databases on a Stardog server. */
+        function list(conn: Connection, params?: Params) : Promise<HTTPBody>;
+        /** Gets number of triples in a database. */
+        function size(conn: Connection, database: string, databaseOptions?: DatabaseOptions, options?: Object, params?: Params) : Promise<HTTPBody>;
+        /** Clears the contents of a database. */
+        function clear(conn: Connection, database: string, databaseOptions?: DatabaseOptions, options?: Object, params?: Params) : Promise<HTTPBody>;
+        /** Gets a mapping of the namespaces used in a database. */
+        function namespaces(conn: Connection, database: string, databaseOptions?: DatabaseOptions, options?: Object, params?: Params) : Promise<HTTPBody>;
+        /** Exports the contents of a database. */
+        function exportData(conn: Connection, database: string, databaseOptions?: DatabaseOptions, options?: Object, params?: Params) : Promise<HTTPBody>;
+
+        /** Database options. */
+        namespace options {
+            /** Gets set of options on a database. */
+            function get(conn: Connection, database: string, params?: Params) : Promise<HTTPBody>;
+            /** Sets options on a database. */
+            function set(conn: Connection, database: string, databaseOptions: DatabaseOptions, params?: Params) : Promise<HTTPMessage>;
+        }
+
+        interface TransactionOptions extends HTTPOptions {
+            encoding: Encoding
+        }
+        type Encoding =
+            'gzip' |
+            'compress' |
+            'deflate' |
+            'identity' |
+            'br';
+        /** Methods for managing transactions in a database. */
+        namespace transaction {
+            /** Begins a new transaction. */
+            function begin(conn: Connection, database: string, params?: Params) : Promise<HTTPBody>;
+            /** Evaluates a SPARQL query within a transaction. */
+            function query(conn: Connection, database: string, transactionId: Guid, q: string, params?: Params) : Promise<HTTPBody>;
+            /** Adds a set of statements to a transaction request. */
+            function add(conn: Connection, database: string, transactionId: Guid, content: string, options: TransactionOptions, params?: Params) : Promise<HTTPBody>;
+            /** Performs a rollback in a given transaction. */
+            function rollback(conn: Connection, database: string, transactionId: Guid, params?: Params) : Promise<HTTPBody>;
+            /** Commits a transaction to the database, removing the transaction and making its changes permanent. */
+            function commit(conn: Connection, database: string, transactionId: Guid, params?: Params) : Promise<HTTPBody>;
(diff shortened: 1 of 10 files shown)


kendall:
I wasn't asked to review this, but I can't help it:

- the content types are mime types, so I don't understand why there's a mime type (type) and a content type (type)...?
- why is `text/plain` in both? It means (?) you can have an HTTP Options object that looks like `{'mimeType': text/plain, 'contentType': 'text/plain'}` which seems odd...
- also HTTP Options is only used (?) as the base class for Transaction Options? If that's true (and will be true for the foreseeable future), it's simpler to just make them be one thing w/out one extending the other? 
- if I'm reading this right (?) an HTTPBody and an HTTPMessage are basically the same thing but body has...a body? Why not just make some body field on Message be optional? It seems cleaner to me for the various functions to all return the same kind of thing, which has (or doesn't) some optional value, rather than 2 different things...

lib/index.d.ts
```diff
@@ -0,0 +1,286 @@
+// Type definitions for stardog.js 0.3.1
+// Project: https://github.com/stardog-union/stardog.js
+// Definitions by: Stephen Nowell <http://github.com/snowell>, Ty Soehngen <http://github.com/tysoeh>
```
arb: Can probably remove. Generally, I think you only add these when the defs don't live with the code and it does in this case.

lib/index.d.ts
```diff
@@ -0,0 +1,286 @@
...
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
```
arb: ✂️

lib/index.d.ts
```diff
@@ -0,0 +1,286 @@
...
+        'text/x-nquads' |
+        'application/trix';
+    
+    type MIMEType =
+        'text/plain' |
+        'application/json' |
+        'text/boolean';
+
+    interface HTTPOptions {
+        headers?: Headers;
```
arb: I'm pretty sure the `headers?` means it's optional... in this case, `headers`, `method` are both required... `method` defaulting to 'GET' if not provided.

arb: That being said, what is this for? Looks like you're using it in transaction stuff. The only exposed options there are "content-type" and "encoding". Don't think you need a whole interface for it.

lib/index.d.ts
```diff
@@ -0,0 +1,286 @@
...
+        method?: string;
+        mimeType?: MIMEType;
+        contentType?: ContentType;
+    }
+
+    interface QueryOptions {
```
arb: Not used?

lib/index.d.ts
```diff
@@ -0,0 +1,286 @@
...
+        query: string;
+        baseURI?: string;
+        limit?: number;
+        offset?: number;
+        reasoning?: boolean;
+    }
+
+    interface HTTPMessage {
```
arb: You can leave this for now, but let's move to just always use HTTPBody and set it to `null` for things that don't really have bodies. The typing on that is going to be object OR boolean OR string OR null which isn't great but is what it is.

lib/index.d.ts
```diff
@@ -0,0 +1,286 @@
...
+        password: string;
+    }
+
+    /** Describes the connection to a running Stardog server. */
+    class Connection {
+        constructor(ConnectionOptions);
+
+        config(options: ConnectionOptions): void;
+        headers(): Headers;
+        uri(resource: Array<string>): string;
```
arb: It's technically not an array of strings but N number of string arguments... not sure how to express that though.

lib/index.d.ts
```diff
@@ -0,0 +1,286 @@
...
+        statusText: string;
+    }
+
+    interface HTTPBody extends HTTPMessage, result {}
+
+    interface result {
+        result: string;
+    }
+
+    interface Params {
```
arb: Do we even need this? Isn't this just `Object` or something?

lib/index.d.ts
```diff
@@ -0,0 +1,286 @@
...
+    
+    namespace server {
+        /** Shuts down a Stardog server. */
+        function shutdown(conn: Connection, params?: Params) : Promise<HTTPMessage>;
+    }
+
+    /** Stardog database actions. */
+    namespace db {
+
+        interface DatabaseOptions {
```
arb: You're not going to be able to type them like this. For the sake of maintenance, I'd just let this be an object and move on. The database object looks like https://github.com/stardog-union/stardog.js/blob/develop/lib/db/dbopts.js

lib/index.d.ts
```diff
@@ -0,0 +1,286 @@
...
+        /** Gets information about a database. */
+        function get(conn: Connection, database: string, params?: Params) : Promise<HTTPBody>;
+        /** Sets a database offline. */
+        function offline(conn: Connection, database: string, params?: Params) : Promise<HTTPBody>;
+        /** Sets a database online. */
+        function online(conn: Connection, database: string, params?: Params) : Promise<HTTPBody>;
+        /** Optimizes a database. */
+        function optimize(conn: Connection, database: string, params?: Params) : Promise<HTTPMessage>;
+        /** Makes a copy of a database. */
+        function copy(conn: Connection, database: string, destination: string, params?: Params) : Promise<HTTPBody>;
```
arb: Can you add comments about variables? I feel like you can do that.

lib/index.d.ts
```diff
@@ -0,0 +1,286 @@
...
+        /** Gets a list of all databases on a Stardog server. */
+        function list(conn: Connection, params?: Params) : Promise<HTTPBody>;
+        /** Gets number of triples in a database. */
+        function size(conn: Connection, database: string, databaseOptions?: DatabaseOptions, options?: Object, params?: Params) : Promise<HTTPBody>;
```
arb: Doesn't match implementation.

lib/index.d.ts
```diff
@@ -0,0 +1,286 @@
...
+        /** Clears the contents of a database. */
+        function clear(conn: Connection, database: string, databaseOptions?: DatabaseOptions, options?: Object, params?: Params) : Promise<HTTPBody>;
```
arb: Doesn't match implementation.

lib/index.d.ts
```diff
@@ -0,0 +1,286 @@
...
+        /** Gets a mapping of the namespaces used in a database. */
+        function namespaces(conn: Connection, database: string, databaseOptions?: DatabaseOptions, options?: Object, params?: Params) : Promise<HTTPBody>;
```
arb: Doesn't match implementation.

lib/index.d.ts
```diff
@@ -0,0 +1,286 @@
...
+        /** Exports the contents of a database. */
+        function exportData(conn: Connection, database: string, databaseOptions?: DatabaseOptions, options?: Object, params?: Params) : Promise<HTTPBody>;
```
arb: Doesn't match implementation.

lib/index.d.ts
```diff
@@ -0,0 +1,286 @@
...
+
+        /** Database options. */
+        namespace options {
+            /** Gets set of options on a database. */
+            function get(conn: Connection, database: string, params?: Params) : Promise<HTTPBody>;
+            /** Sets options on a database. */
+            function set(conn: Connection, database: string, databaseOptions: DatabaseOptions, params?: Params) : Promise<HTTPMessage>;
```
arb: Same as above re `DatabaseOptions`

lib/index.d.ts
```diff
@@ -0,0 +1,286 @@
...
+        type Encoding =
+            'gzip' |
+            'compress' |
+            'deflate' |
+            'identity' |
+            'br';
+        /** Methods for managing transactions in a database. */
+        namespace transaction {
+            /** Begins a new transaction. */
+            function begin(conn: Connection, database: string, params?: Params) : Promise<HTTPBody>;
```
arb: All of the transaction promises also attach `transactionId` to the resolved body to make chaining a little easier.

lib/index.d.ts
```diff
@@ -0,0 +1,286 @@
...
+            /** Evaluates a SPARQL query within a transaction. */
+            function query(conn: Connection, database: string, transactionId: Guid, q: string, params?: Params) : Promise<HTTPBody>;
```
arb: `q` = `query`. The function argument names don't have to match the implementation, just the order. Remember, at the end of the day this is for humans first and foremost.

lib/index.d.ts
```diff
@@ -0,0 +1,286 @@
...
+
+        /** Database options. */
+        namespace options {
+            /** Gets set of options on a database. */
+            function get(conn: Connection, database: string, params?: Params) : Promise<HTTPBody>;
+            /** Sets options on a database. */
+            function set(conn: Connection, database: string, databaseOptions: DatabaseOptions, params?: Params) : Promise<HTTPMessage>;
+        }
+
+        interface TransactionOptions extends HTTPOptions {
```
arb: Probably just `contentType` and `encoding` here.

lib/index.d.ts
```diff
@@ -0,0 +1,286 @@
...
+            /** Removes a set of statements from a transaction request. */
+            function remove(conn: Connection, database: string, transactionId: Guid, content: string, options: TransactionOptions, params?: Params) : Promise<HTTPBody>;
+        }
+
+        /** Methods for managing integrity constraints in a database. */
+        namespace icv {
+            /** Gets the set of integrity constraints on a given database. */
+            function get(conn: Connection, database: string, options: Object, params?: Params) : Promise<HTTPBody>;
+            /** Sets a new set of integrity constraints on a given database. */
+            function set(conn: Connection, database: string, icvAxioms: string, options: Object, params?: Params) : Promise<HTTPMessage>;
```
arb: Currently only valid key on `options` is `contentType`.

lib/index.d.ts
```diff
@@ -0,0 +1,286 @@
...
+            /** Removes all integrity constraints from a given database. */
+            function clear(conn: Connection, database: string, options: Object, params?: Params) : Promise<HTTPMessage>;
+            /** Converts a set of integrity constraints into an equivalent SPARQL query for a given database. */
+            function convert(conn: Connection, database: string, icvAxioms: string, options: Object, params?: Params) : Promise<HTTPBody>;
```
arb: Same as above.

lib/index.d.ts
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    interface PropertyOptions {
+        uri: string,
+        property: string
+    }
+    /** Query actions to perform on a database. */
+    namespace query {
+        /** Gets the values for a specific property of a URI individual. */
+        function property(conn: Connection, database: string, options?: PropertyOptions, params?: Params) : result;
```
arb: `options` is not optional. Should be renamed `config` in the type defs.

arb: `result` is still a promise.

lib/index.d.ts
```diff
@@ -0,0 +1,286 @@
...
+        /** Kills an actively running query. */
+        function kill(conn: Connection, queryId: string) : Promise<HTTPMessage>;
+        /** Gets information about an actively running query. */
+        function get(conn: Connection, queryId: string) : Promise<HTTPBody>;
+
+        interface StoredQueryOptions {
+            name: string,
+            database: string,
+            query: string,
+            shared: boolean
```
arb: Is there a way to indicate what something defaults to? I feel like there is...

lib/index.d.ts
```diff
@@ -0,0 +1,286 @@
...
+        }
+        /** Manages stored queries. */
+        namespace stored {
+            /** Stores a query in Stardog, either on the system level or for a given database. */
+            function create(conn: Connection, storedQuery: StoredQueryOptions, params?: Params) : Promise<HTTPBody>
```
arb: Please correct me if I'm wrong here, but I don't think the names of the arguments have to line up, just their positions. In that case, this could probably have a better name.

lib/index.d.ts
```diff
@@ -0,0 +1,286 @@
...
+        }
+    }
+
+    /** Administrative actions for managing users, roles, and their permissions. */
+    namespace user {
+
+        interface User {
+            username: string;
+            password: string;
+            superuser?: boolean;
```
arb: Pretty sure this isn't optional.

arb:
@kendall 
- our thinking was there is a list of mime types that the server cares about, some of them can be used as accept headers, others can be used as content type headers. It is not documented that all of them can be used for either header, so each list of mime-types are scoped to just the ones the documentation talks about.
- agreed on the response body thing. We've talked about always using the same response thing so the shape is always, roughly, similar.

kendall:
Okay in re: mimetypes that makes more sense, but it's very opaque as-is. How explicitly AcceptMimeTypes and ResponseMimeTypes or something that's very obvious what distinguishes the two groups? Or just one list of types where each type has some kind of tag (or more than one) about which role it can play. I don't know how or whether you can do it that way in this type system.

arb:
Yeah, let's go for the verbose option. The single list with tags I think would end up being more confusing than helpful (assuming it's even possible).

@tysoeh make it so!

PR #92, opened 2017-07-18
✨ - Expose update endpoint

### LINKED ISSUE #88: SPARQL Updates
[SPARQL updates](https://www.stardog.com/docs/#_sparql_update_2

arb:
Closes #88

Adds logic to handle update type queries. Users should be able to just call
query.execute regardless of what it is with this change.

### DIFF
--- a/lib/query/main.js
+++ b/lib/query/main.js
@@ -3,7 +3,7 @@ const lodashGet = require('lodash/get');
 const qs = require('querystring');
 
 const { httpBody, httpMessage } = require('../response-transforms');
-const { mimeType, buildQuery } = require('./utils');
+const { mimeType, buildQuery, queryType } = require('./utils');
 
 const dispatchQuery = (conn, config, params) => {
   const headers = conn.headers();
@@ -12,23 +12,26 @@ const dispatchQuery = (conn, config, params) => {
 
   const body = buildQuery(config.query, params);
 
-  return fetch(conn.uri(config.database, 'query'), {
+  return fetch(conn.uri(config.database, config.resource), {
     method: 'POST',
     body: qs.stringify(body),
     headers,
   }).then(httpBody);
 };
 
-const execute = (conn, database, query, params) =>
-  dispatchQuery(
+const execute = (conn, database, query, params) => {
+  const type = queryType(query);
+  return dispatchQuery(
     conn,
     {
       database,
       query,
       accept: mimeType(query),
+      resource: type === 'update' ? 'update' : 'query',
     },
     params
   );
+};
 
 const property = (conn, database, options, params) =>
   execute(
--- a/lib/query/utils.js
+++ b/lib/query/utils.js
@@ -26,6 +26,21 @@ const queryType = query => {
     return 'describe';
   }
 
+  if (
+    q.startsWith('insert') ||
+    q.startsWith('delete') ||
+    q.startsWith('with') ||
+    q.startsWith('load') ||
+    q.startsWith('clear') ||
+    q.startsWith('create') ||
+    q.startsWith('drop') ||
+    q.startsWith('copy') ||
+    q.startsWith('move') ||
+    q.startsWith('add')
+  ) {
+    return 'update';
+  }
+
   return null;
 };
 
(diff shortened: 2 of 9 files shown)


lib/query/main.js
```diff
@@ -12,23 +12,26 @@ const dispatchQuery = (conn, config, params) => {
...
-  dispatchQuery(
+const execute = (conn, database, query, params) => {
+  const type = queryType(query);
+  return dispatchQuery(
     conn,
     {
       database,
       query,
       accept: mimeType(query),
+      resource: type === 'update' ? 'update' : 'query',
```
snowell: Why not have queryType(query) return a default value instead of this?

lib/query/utils.js
```diff
@@ -36,7 +51,7 @@ const mimeType = query => {
     return 'application/sparql-results+json';
   }
 
-  if (type === 'ask') {
+  if (type === 'ask' || type === 'update') {
     return 'text/boolean';
```
snowell: Technically speaking, updates don't return boolean. But this probably doesn't make a difference here

lib/query/utils.js
```diff
@@ -26,6 +26,21 @@ const queryType = query => {
     return 'describe';
   }
 
+  if (
```
snowell: Unless you're stripping out prefix declarations at the beginning of the query, you're not always going to hit one of these.

Example:

```
prefix foo: <http://messing.up.your/logic#>
insert data { foo:foo foo:trololo 42 }
```

test/getDB.spec.js
```diff
@@ -21,7 +21,7 @@ describe('getDB()', () => {
 
   it('A response of the DB info should not be empty', () =>
     db.get(conn, database).then(res => {
-      expect(res.body.length).toBe(7482);
+      expect(res.body.length).toBeGreaterThan(7400);
```
snowell: Why the magic number? "Not empty" is really >0

test/query.spec.js
```diff
@@ -193,4 +193,49 @@ describe('query.execute()', () => {
...
+        .then(res => {
+          expect(res.status).toBe(200);
+          expect(res.body).toBe(null);
+          return execute(`select * {:foo ?p ?o}`);
+        })
+        .then(res => {
+          expect(res.body.results.bindings).toHaveLength(1);
+        }));
+
+    it('should remove some data', () =>
```
snowell: This test is asserting that a triple is not in the database, attempting to delete it twice, and then asserting that it's still not there. I think something got lost in translation

test/query.spec.js
```diff
@@ -193,4 +193,49 @@ describe('query.execute()', () => {
...
+      execute(`insert {?s :someProp 0} where {?s :bar :baz}`)
+        .then(() =>
+          execute(`
+          delete {?s :someProp ?o}
+          insert {?s :someProp 42}
+          where {?s :someProp ?o}
+        `)
+        )
+        .then(res => {
+          expect(res.status).toBe(200);
```
snowell: Since we're making assertions that the server did the right thing, we should assert that the {?s :someProp 42} triple is now in the db. Maybe even that the {?s :someProp 0} one isn't.

test/getDB.spec.js
tysoeh: This is intended in the commit?

arb: Yes.

lib/query/utils.js
arb: I am. Line 4-11.

test/query.spec.js
```diff
@@ -193,4 +193,49 @@ describe('query.execute()', () => {
...
+          return execute('delete where {?s :someProp ?o}');
```
tysoeh: No assertion after this?

lib/query/main.js
arb: I don't understand the question. The type is what it is, I'm just computing what endpoint to hit based on the `type`. 

Can you clarify the question?

snowell: Ah nope, it was I who misunderstood what was happening

test/query.spec.js
arb: Didn't think it was necessary. We've already asserted deletes work.

test/query.spec.js
arb: Yeah, I really hosed that up. Good call out.

test/query.spec.js
arb: Ok, I added a few extra assertions to this test to cover that case.

PR #93, opened 2017-07-19
Cleanup typedefs

tysoeh:
Now has four top-level namespaces (`db`, `query`, `user`, `server`) and one class (`Connection`) instead of a global `stardog` namespace.

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -1,11 +1,30 @@
 // TypeScript Version: 2.1
 
 import Headers from 'fetch-ponyfill';
-
-declare global {
   
-  /** stardog.js: The Stardog JS API*/
-  namespace Stardog {
+declare const db: stardog.db;
+declare const query: stardog.query;
+declare const server: stardog.server;
+declare const user: stardog.user;
+
+    
+interface ConnectionOptions {
+    endpoint: string;
+    username: string;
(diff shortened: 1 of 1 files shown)


lib/index.d.ts
```diff
@@ -1,252 +1,257 @@
...
         status: string;
         statusText: string;
-        result: Object | string | boolean | null;
-    }
-    
-    interface ConnectionOptions {
-        endpoint: string;
-        username: string;
-        password: string;
+        result: object | string | boolean | null;
```
tysoeh: Pretty sure this needs to be `Object`

lib/index.d.ts
```diff
@@ -1,252 +1,257 @@
...
+/** stardog.js: The Stardog JS API*/
 
-declare global {
-  
-  /** stardog.js: The Stardog JS API*/
-  namespace Stardog {
+import Headers from 'fetch-ponyfill';
 
-    type ContentMimeTypes =
+namespace HTTP {
```
tysoeh: `declare` is required for top level declarations

lib/index.d.ts
arb: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html

PR #98, opened 2017-07-24
generate docs from typescript decl file

tysoeh:


lib/index.d.ts
```diff
@@ -40,7 +46,7 @@ declare namespace Stardog {
 
     /** Describes the connection to a running Stardog server. */
     export class Connection {
-        constructor(ConnectionOptions);
+        constructor(options: ConnectionOptions);
```
arb: Good catch there.

scripts/docs.js
```diff
@@ -0,0 +1,144 @@
...
+const getPath = obj => {
+  if (obj.parent === null || obj.parent.name === `"${pathToDeclaration}"`)
+    return obj.name;
+  return `${getPath(obj.parent)}.${obj.name}`;
+};
+
+const getMarkdown = (obj, parent = null) => {
+  const { kind } = obj;
+  /* FunctionDeclaration */
+  if (kind === 216) {
```
arb: typedocs really doesn't export some kind of enum?

scripts/docs.js
```diff
@@ -0,0 +1,144 @@
...
+      `Expects the following parameters:\n\n${obj.parameters.reduce(
+        (acc, val) => acc.concat(`- ${val.name} (${getType(val.type)})\n\n`),
+        ''
+      )}`,
+      `Returns [${type}](#${typeHeading})`,
+      '\n\n'
+    );
+  }
+  /* ClassDeclaration */
+  if (kind === 217) {
```
arb: It might make sense to break these different `kind` checks out into their own functions.

scripts/docs.js
```diff
@@ -0,0 +1,144 @@
...
+      obj.members.forEach(child => {
+        md = md.concat(getMarkdown(child, obj));
+      });
+    }
+    return md;
+  }
+  return '';
+};
+
+Promise.resolve(typedocs.generate([pathToDeclaration]))
```
arb: Is `tydocs.generate` already a promise? If so, you can skip `Promise.resolve`.

scripts/docs.js
```diff
@@ -0,0 +1,144 @@
...
+  .then(res => {
+    let markdown = '';
+    res.forEach(obj => {
+      markdown = markdown.concat(getMarkdown(obj));
+    });
+    return markdown;
+  })
+  .then(markdown => {
+    fs.readFile(pathToREADME, 'utf8', (err, data) => {
+      data = data.split('# API');
```
arb: So what I've seen other people do is like put some kind of comment in the markdown

`<-- API GOES HERE -->`

or something to prevent any potential collisions.

https://gist.github.com/jonikarppinen/47dc8c1d7ab7e911f4c9

(Example of markdown comments)

scripts/docs.js
```diff
@@ -0,0 +1,144 @@
...
+  )
+  .then(res => {
+    let markdown = '';
+    res.forEach(obj => {
+      markdown = markdown.concat(getMarkdown(obj));
+    });
+    return markdown;
+  })
+  .then(markdown => {
+    fs.readFile(pathToREADME, 'utf8', (err, data) => {
```
arb: You should return a new promise here wrapping `fs.writeFile`. If you don't then the promise chain is over before this operation is done and it's unlikely to be caught in your `catch` block.

scripts/docs.js
```diff
@@ -0,0 +1,144 @@
...
+    type.startsWith('string') ||
+    type.startsWith('Object') ||
+    type.startsWith('array') ||
+    type.startsWith('boolean')
+  ) {
+    return `\`${type}\``;
+  }
+  if (type.startsWith('{')) {
+    const customType = type.slice(
+      type.indexOf(': ') + 2,
```
arb: I think you could do `split` here maybe

scripts/docs.js
```diff
@@ -0,0 +1,144 @@
...
+        ''
+      )}`,
+      `Returns [${type}](#${typeHeading})`,
+      '\n\n'
+    );
+  }
+  /* ClassDeclaration */
+  if (kind === 217) {
+    let md = `## Class <a name="${obj.name}">${obj.name}</a>\n\n`;
+    const constructor = obj.members.shift().parameters[0];
```
arb: Should probably call this something else... `constructor` is a kind of a reserved word.

scripts/docs.js
```diff
@@ -0,0 +1,144 @@
...
+    if (obj.members) {
+      obj.members.forEach(child => {
+        md = md.concat(getMarkdown(child));
+      });
+    }
+    return md;
+  }
+  /* InterfaceDeclaration */
+  if (kind === 218) {
+    return ''.concat(
```
arb: Why the extra `''.concat`?

README.md
```diff
@@ -76,3 +76,959 @@ query.execute(conn, 'select distinct ?s where { ?s ?p ?o }', {
...
+### <a name="Encodings">Encodings</a>
+
+One of the following values:
+
+`'gzip' |
+                'compress' |
+                'deflate' |
+                'identity' |
+                'br'`
+### <a name="TransactionResponse">TransactionResponse</a>
```
arb: This seems to be missing things. A transaction response has everything http.body has plus a transaction Id.

README.md
```diff
@@ -76,3 +76,959 @@ query.execute(conn, 'select distinct ?s where { ?s ?p ?o }', {
...
+
+Object with the following values:
+
+- transactionId (`string`)
+
+### <a name="TransactionOptions">TransactionOptions</a>
```
arb: Also, why do some of these anchor tags have names and others not? Also if they must have a name, I'd prefer if it were all lower cased.

README.md
```diff
@@ -76,3 +76,959 @@ query.execute(conn, 'select distinct ?s where { ?s ?p ?o }', {
...
+
+Object with the following values:
+
+- contentType (`HTTP.ContentMimeTypes`)
+- encoding (`Encodings`)
+
+### `db.transaction.begin(conn, database, params)`
```
arb: Let's do something to keep these headers consistent. Either use full anchor tags or the `###` not both.

scripts/docs.js
tysoeh: what do you mean when you say "extra"? I'm just using `''.concat` because it reads so much cleaner than splitting the markdown on either side of the `.concat`.

scripts/docs.js
tysoeh: it's not

scripts/docs.js
arb: This still isn't right. This function returns `undefined` as is. Line 151 should start with `return new Promise(resolve, reject) {...}`

PR #99, opened 2017-07-25
✅ - Add tests for update SPARQL queries. Fix #91

### LINKED ISSUE #91: More SPARQL Update Tests
Flesh out more tests to exercise the SPARQL update statements.

snowell:


test/query.spec.js
```diff
@@ -195,42 +195,50 @@ describe('query.execute()', () => {
   });
 
   describe('update', () => {
-    it('should insert some data', () =>
-      execute(`insert data {:foo :bar :baz}`)
+    let clearDatabase = () => execute('clear all');
```
arb: `const`. Also "clear all" is a real query?

arb: Actually you don't even need to make this a variable, just do `afterEach(() => execue('clear all'))`

test/query.spec.js
```diff
@@ -195,42 +195,50 @@ describe('query.execute()', () => {
...
+    afterEach(clearDatabase);
+
+    it('should support "insert data"', () => {
+      let data = '{:foo :bar :baz}';
```
arb: `const` or just inline it with the query.

test/query.spec.js
```diff
@@ -195,42 +195,50 @@ describe('query.execute()', () => {
...
           expect(res.body.results.bindings).toHaveLength(1);
-        }));
+        });
+    });
 
-    it('should remove some data', () =>
-      execute(`select * {<urn:foo> ?p ?o}`)
+    it('should support "delete data"', () => {
+      let data = '<urn:foo> <urn:bar> <urn:baz>';
+      let select = `select * {<urn:foo> ?p ?o}`;
```
arb: Same as above and everywhere else you do this pattern.

test/query.spec.js
snowell: It is! https://www.w3.org/TR/sparql11-update/#clear

PR #100, opened 2017-07-25
📝 - Add a bunch of JSDoc for function parameters. Fix #97

### LINKED ISSUE #97: Function parameter documentation
https://blogs.msdn.microsoft.com/typescript/2013/01/21/announcing-typescript-0-8-2/

Add some human words for what function parameters are.

snowell:
Also made incredibly minor changes like removing parameters that weren't used

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -50,7 +50,7 @@ declare namespace Stardog {
     /** Stardog HTTP server actions. */
     export namespace server {
         /** Shuts down a Stardog server. */
-        function shutdown(conn: Connection, params?: Object): Promise<HTTP.Body>;
+        function shutdown(conn: Connection, params?: object): Promise<HTTP.Body>;
     }
 
     /** Stardog database actions. */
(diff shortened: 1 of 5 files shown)


tysoeh:
@arb the `@see`s get appended to the generic documentation. the `@param`s are extracted and attached to the parameter objects. so for example, with these param comments, the doc description for `db.create` would look like

```
db.create(conn, database, databaseOptions, options, params)

Creates a new database.

@see {@link https://github.com/stardog-union/stardog.js/blob/master/lib/db/dbopts.js} @see {@link https://www.stardog.com/docs/#_database_admin}

Expects the following parameters:
...
```

I'll try to see if there's a way to attach links to the params in a typescript decl file before approving

lib/index.d.ts
```diff
@@ -23,7 +23,7 @@ declare namespace Stardog {
         export interface Body {
             status: string;
             statusText: string;
-            result: object | string | boolean | null;
+            result: Object | string | boolean | null;
```
arb: it's `object, lowercase "o". We had this change before.


> Don’t ever use the types Number, String, Boolean, or Object. These types refer to non-primitive boxed objects that are almost never used appropriately in JavaScript code.

-Taken from [here](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html)

lib/index.d.ts
```diff
@@ -58,111 +58,111 @@ declare namespace Stardog {
...
          * Creates a new database.
          * 
-         * @param conn the Stardog server connection
-         * @param database the name of the database to create
-         * @param databaseOptions an object specifying the various metadata options for this database. @see {@link https://github.com/stardog-union/stardog.js/blob/master/lib/db/dbopts.js} @see {@link https://www.stardog.com/docs/#_database_admin}
-         * @param options an object specifying a list of RDF files to bulk load into the database at creation time
-         * @param params additional parameters if needed
+         * @param {Connection} conn the Stardog server connection
+         * @param {string} database the name of the database to create
+         * @param {Object} databaseOptions an object specifying the various metadata options for this database. See: https://github.com/stardog-union/stardog.js/blob/master/lib/db/dbopts.js
```
arb: Lowercase "o"

Issue #85, opened 2017-07-13
Missing APIs


arb:
## List of known missing endpoints

- [x] get list of stored procedures
- [x] remove specified set of ICV axioms
- [x] ICV isValid check
- [x] ICV get all ViolationBindings
- [x] ICV get specific ViolationBindings
- [x] ~~[SPARQL updates](https://www.stardog.com/docs/#_sparql_update_2)~~
- [ ] ~~List of stored procedures~~
- [x] reasoning explain
- [x] removeICV

tysoeh:
[icv methods](https://github.com/stardog-union/stardog/blob/develop/icv/api/main/src/com/complexible/stardog/icv/ICValidator.java)

rivettp:
SPARQL update queries (see #73)

PR #102, opened 2017-07-27
✨ - Add reasoning endpoints and tests as part of #85

snowell:


lib/response-transforms.js
```diff
@@ -32,4 +32,8 @@ module.exports = {
       });
     }
   },
+  jsonify: res => {
```
arb: Just do this inline where we need it. Also why do we even need to do this? Does the server not respond with JSON? Have you set the correct "accept" header?

lib/db/reasoning.js
```diff
@@ -0,0 +1,120 @@
...
+  headers.append('Accept', 'text/boolean');
+
+  const uri = `${conn.uri(
+    database,
+    'reasoning',
+    'consistency'
+  )}${options.namedGraph ? `?graph-uri=${options.namedGraph}` : ''}`;
+
+  return fetch(uri, {
+    method: 'GET',
```
arb: Implied.

lib/db/reasoning.js
```diff
@@ -0,0 +1,120 @@
...
+  inference,
+  options,
+  params = {}
+) => {
+  const headers = conn.headers();
+  headers.append('Content-Type', options.contentType);
+  if (options.encoding) {
+    headers.append('Content-Encoding', options.encoding);
+  }
+  console.info(conn.uri(database, 'reasoning', transactionId, 'explain'));
```
arb: ✂️

lib/db/reasoning.js
```diff
@@ -0,0 +1,120 @@
...
+
+const explainInconsistencyInTx = (
+  conn,
+  database,
+  transactionId,
+  options,
+  params = {}
+) => {
+  const headers = conn.headers();
+  console.info(
```
arb: ✂️

lib/db/reasoning.js
```diff
@@ -0,0 +1,120 @@
...
+    'explain',
+    'inconsistency'
+  )}${options.namedGraph ? `?graph-uri=${options.namedGraph}` : ''}`;
+
+  return fetch(uri, {
+    method: 'POST',
+    headers,
+  })
+    .then(httpBody)
+    .then(attachTransactionId(transactionId));
```
arb: Where is `attachTransactionId` defined?

test/reasoning.spec.js
```diff
@@ -0,0 +1,99 @@
...
+  ConnectionFactory,
+} = require('./setup-database');
+
+const { reasoning, transaction } = db;
+
+describe('reasoning commands', () => {
+  const database = generateDatabaseName();
+  const conn = ConnectionFactory();
+
+  const beginTx = transaction.begin.bind(null, conn, database);
```
arb: You liked that trick eh? 😄

test/reasoning.spec.js
```diff
@@ -0,0 +1,99 @@
...
+  const commitTx = transaction.commit.bind(null, conn, database);
+
+  beforeAll(seedDatabase(database));
+  afterAll(dropDatabase(database));
+
+  it('should be able to check consistency', () => {
+    return reasoning.consistency(conn, database, {}).then(res => {
+      expect(res.status).toBe(200);
+      expect(res.body).toEqual('true');
```
arb: This should be `true`, not `"true"` the string "true" is less than useless.

test/reasoning.spec.js
```diff
@@ -0,0 +1,99 @@
...
+  });
+
+  it('should explain inferences', () => {
+    return reasoning
+      .explainInference(conn, database, '<urn:A> a <urn:B> .', {
+        contentType: 'text/turtle',
+      })
+      .then(res => {
+        expect(res.status).toBe(200);
+        expect(res.body.proofs).toBeTruthy();
```
arb: Eh... this is weak. Can we do any better?

test/reasoning.spec.js
```diff
@@ -0,0 +1,99 @@
...
+      });
+  });
+
+  it('should explain inconsistency', () => {
+    return reasoning.explainInconsistency(conn, database, {}).then(res => {
+      expect(res.status).toBe(200);
+      expect(res.body.proofs).toBeTruthy();
```
arb: Same as above.

test/reasoning.spec.js
```diff
@@ -0,0 +1,99 @@
...
+          conn,
+          database,
+          res.transactionId,
+          '<urn:A> a <urn:B> .',
+          { contentType: 'text/turtle' }
+        );
+      })
+      .then(res => {
+        expect(res.status).toBe(200);
+        expect(res.body.proofs).toBeTruthy();
```
arb: Same as above.

test/reasoning.spec.js
```diff
@@ -0,0 +1,99 @@
...
+        // Makeshift 'finally'
+        return res;
+      })
+      .catch(res => res)
```
arb: Please explain this... why do we need a `catch`? It's rare and likely an anti-pattern to have a `.catch` followed by a `.then` at the same async nesting level.

test/reasoning.spec.js
```diff
@@ -0,0 +1,99 @@
...
+      .then(res => commitTx(res.transactionId));
+  });
+
+  it('should successfully get the schema', () => {
+    return reasoning.schema(conn, database).then(res => {
+      expect(res.status).toBe(200);
+      expect(res.body).not.toEqual('');
```
arb: Is there a regex this could possibly match?

lib/db/reasoning.js
snowell: Ugh, thought I got rid of all of these

lib/response-transforms.js
snowell: Accept header is irrelevant. The server isn't setting content-type, so when it gets to us here it's considered plain text. I'd originally fixed the server, but eventually settled on testing the behavior that IS instead of the behavior that SHOULD be.

snowell: Factored it out into its own function so that the promise chains would look cleaner, since we do it more than once and there's every likelihood to expect to have to do it more as things are added if they don't return json

lib/db/reasoning.js
snowell: Nowhere. Bad copy/paste error. I'd originally had this in transaction.js

test/reasoning.spec.js
snowell: It asserts that the response body came back with a "proofs" key. I don't particularly care what the value is

test/reasoning.spec.js
snowell: The method returns a block of RDF, the contents of which I can't be sure are guaranteed. As long as it's not empty, I'm satisfied that it worked

test/reasoning.spec.js
snowell: A bit of a hack I found at https://stackoverflow.com/questions/35999072/what-is-the-equivalent-of-bluebird-promise-finally-in-native-es6-promises to hold over until ES2018 when native Promise gets a `finally` equivalent. Opening a tx without closing it is no bueno, so having the `then` after a `catch` ensures that it will be called no matter what.

lib/response-transforms.js
arb: What I'm saying is, if you set the accept header in the request, it's possible the server will send back the corresponding content-type header?

snowell: It's a good thought, but I tried it, and no. It would require a server change.

lib/index.d.ts
```diff
@@ -133,6 +133,62 @@ declare namespace Stardog {
...
+
+            /**
+             * Provides the reason why a database is inconsistent, as reported by db.reasoning.consistency
+             * 
+             * @param {Connection} conn the Stardog server connection
+             * @param {string} database the name of the database
+             * @param {object} options an object optionally specifying the URI of a graph to evaluate
+             * @param {object} params additional parameters if needed  
+             */
+            function explainInconsistency(conn: Connection, database: string, options: { namedGraph: string }, params?: object);
```
tysoeh: `options` should be `options?` I believe

lib/index.d.ts
```diff
@@ -133,6 +133,62 @@ declare namespace Stardog {
...
+            /**
+             * Provides the reason why a database is inconsistent, as reported by db.reasoning.consistency
+             * 
+             * @param {Connection} conn the Stardog server connection
+             * @param {string} database the name of the database
+             * @param {string} transactionId: the UUID of the transaction as returned by db.transaction.begin
+             * @param {object} options an object optionally specifying the URI of a graph to evaluate
+             * @param {object} params additional parameters if needed  
+             */
+            function explainInconsistencyInTx(conn: Connection, database: string, transactionId: string, options: { namedGraph: string }, params?: object);
```
tysoeh: here as well

lib/index.d.ts
snowell: I was under the impression that you couldn't have more than 1 optional parameter, because how would you pass in just the second one, for example?

tysoeh: good point- I suppose that's true. if so, we could still do `options: { namedGraph?: string }`

lib/response-transforms.js
arb: Ok but still... I don't want this to become part of the public API, so can we just add this in the file where it's needed and not to the global HTTP response manipulation functions.

lib/index.d.ts
arb: You'd do `explainInconsistency(conn, 'db', undefined, { params})`

test/reasoning.spec.js
```diff
@@ -0,0 +1,99 @@
+/* eslint-env jest */
+
+const fs = require('fs');
```
tysoeh: this import is unused and is throwing an eslint error which breaks the build

lib/db/reasoning.js
```diff
@@ -0,0 +1,117 @@
+const { fetch } = require('../fetch');
+
+const { httpBody } = require('../response-transforms');
+
+const jsonify = res => {
+  res.headers.append('content-type', 'application/json');
+  return res;
+};
+
+const consistency = (conn, database, options, params) => {
```
arb: I think you need `options = {}` here. If we are saying this argument is optional, and it is, then you shouldn't have to pass `{}` to keep the function from crashing.

lib/db/reasoning.js
```diff
@@ -0,0 +1,117 @@
...
+    'consistency'
+  )}${options.namedGraph ? `?graph-uri=${options.namedGraph}` : ''}`;
+
+  return fetch(uri, {
+    headers,
+  }).then(httpBody);
+};
+
+// contentType - application/x-turtle, text/turtle, application/rdf+xml, text/plain, application/x-trig, text/x-nquads, application/trix
+const explainInference = (conn, database, inference, options, params = {}) => {
```
arb: So since `options` is required here, because you need to specify contentType, I've been calling those kinds of arguments, `config`.

lib/db/reasoning.js
```diff
@@ -0,0 +1,117 @@
...
+  return fetch(conn.uri(database, 'reasoning', 'explain'), {
+    method: 'POST',
+    headers,
+    body: inference,
+  })
+    .then(jsonify)
+    .then(httpBody);
+};
+
+const explainInconsistency = (conn, database, options, params = {}) => {
```
arb: `options = {}`

lib/db/reasoning.js
```diff
@@ -0,0 +1,117 @@
...
+    .then(httpBody);
+};
+
+// contentType - application/x-turtle, text/turtle, application/rdf+xml, text/plain, application/x-trig, text/x-nquads, application/trix
+const explainInferenceInTx = (
+  conn,
+  database,
+  transactionId,
+  inference,
+  options,
```
arb: Same as above, this should be `config` since it seems to be required.

lib/db/reasoning.js
```diff
@@ -0,0 +1,117 @@
...
+    headers,
+    body: inference,
+  }).then(httpBody);
+};
+
+const explainInconsistencyInTx = (
+  conn,
+  database,
+  transactionId,
+  options,
```
arb: `options = {}`

PR #103, opened 2017-07-28
✨ - Flesh out ICV endpoint and consolidate its tests

snowell:


lib/db/icv.js
```diff
@@ -5,16 +5,16 @@ const { httpBody } = require('../response-transforms');
...
   return fetch(conn.uri(database, 'icv'), {
     headers,
   }).then(httpBody);
 };
 
-const set = (conn, database, icvAxioms, options, params = {}) => {
+const add = (conn, database, icvAxioms, options, params = {}) => {
   const headers = conn.headers();
-  headers.append('Content-Type', options.contentType || 'text/plain');
+  headers.append('content-type', options.contentType || 'text/plain');
```
arb: We've been doing uppercase.

lib/db/icv.js
```diff
@@ -53,9 +63,120 @@ const convert = (conn, database, icvAxiom, options, params = {}) => {
...
+  }
+  const query = qs.stringify(queryParams);
+  const uri = `${conn.uri(
+    database,
+    'icv',
+    transactionId,
+    'validate'
+  )}${query.length > 0 ? `?${query}` : ''}`;
+
+  console.info(uri);
```
arb: ✂️

lib/index.d.ts
```diff
@@ -269,15 +269,26 @@ declare namespace Stardog {
...
+            function add(conn: Connection, database: string, icvAxioms: string, options?: { contentType: HTTP.ContentMimeTypes }, params?: object): Promise<HTTP.Body>;
+
+            /** 
+             * Removes integrity constraints from a given database. 
+             * 
+             * @param {Connection} conn the Stardog server connection
+             * @param {string} database the name of the database
+             * @param {string} icvAxioms an RDF block containing the axioms to be removed
+             * @param {object} options an object specifying the contentType of the RDF data. Default: text/turtle
+             * @param {object} params additional parameters if needed
```
arb: You may want to add a note about the `contentType` has to "match" or make sense or whatever with regard to the `icvAxioms` parameter.

test/icv.spec.js
```diff
@@ -0,0 +1,117 @@
...
+  it('should convert constraint axioms to a SPARQL query', () =>
+    icv
+      .convert(conn, database, icvAxioms, { contentType: 'text/turtle' })
+      .then(res => {
+        expect(res.body.startsWith('SELECT')).toBe(true);
+      }));
+
+  // Server is incorrectly returning text/plain. Once that fix is live
+  // we need to update these tests to expect the boolean
+  it('should validate constraints', () =>
```
arb: Did you try setting the accept header to "text/boolean"? You probably already did but just wanted to check.

test/icv.spec.js
```diff
@@ -0,0 +1,117 @@
...
+      .then(res => {
+        expect(res.status).toBe(200);
+        // expect(res.body).toBe(false);
+        expect(res.body).toBe('false');
+      }));
+
+  it('should report violations', () =>
+    icv.violations(conn, database, '', {}).then(res => {
+      expect(res.status).toBe(200);
+      expect(res.body).toBeNull();
```
arb: Really? That's kind of weird to have a 200 and a `null` body? Is the actual result `""`?

lib/db/icv.js
```diff
@@ -53,9 +63,120 @@ const convert = (conn, database, icvAxiom, options, params = {}) => {
...
+    headers,
+  }).then(httpBody);
+};
+
+const violationsInTx = (
+  conn,
+  database,
+  transactionId,
+  constraints,
+  options,
```
arb: I'd make this `= {}` as well. As it is, even if you want the default `contentType` you'd have to pass in an empty object which is kind of confusing for the user.

lib/db/icv.js
```diff
@@ -53,9 +63,120 @@ const convert = (conn, database, icvAxiom, options, params = {}) => {
...
+    : ''}`;
+
+  return fetch(uri, {
+    method: 'POST',
+    body: constraints,
+    headers,
+  }).then(httpBody);
+};
+
+const violationsInTx = (
```
arb: I'd argue, not strongly mind you, that this should be under the transaction namespace and not under the icv namespace.

snowell: I would agree, except I attempted to do it that way first and the tests seemed to have trouble finding the right namespaces or something. But doing it this way worked, so I went with it. Will re-investigate

test/icv.spec.js
snowell: I did. The server ignores it. Issue 4080 set up to track it; unsure of what release the final fix may make it into

test/icv.spec.js
snowell: It is indeed. Because the response is multipart/mixed, so if there are zero parts, you get a null body. Content-Length: 0

lib/db/icv.js
```diff
@@ -119,8 +117,8 @@ const validateInTx = (
 
 const violations = (conn, database, constraints, options, params = {}) => {
```
arb: `options = {}`

PR #105, opened 2017-08-01
get single user info

tysoeh:
adds user.get which retrieves all information on a given user

### DIFF
--- a/test/getUser.spec.js
+++ b/test/getUser.spec.js
@@ -1,0 +1,36 @@
+/* eslint-env jest */
+
+const { user } = require('../lib');
+const { generateRandomString, ConnectionFactory } = require('./setup-database');
+
+const { role } = user;
+
+describe('user.get', () => {
+  const conn = ConnectionFactory();
+
+  it('should return information on a given user', () => {
+    const r = generateRandomString();
+    const permission = {
+      action: 'READ',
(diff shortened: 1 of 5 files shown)


test/getUser.spec.js
```diff
@@ -0,0 +1,40 @@
...
+      .create(conn, {
+        name: r,
+      })
+      .then(() => user.setRoles(conn, 'anonymous', [r]))
+      .then(() => user.assignPermission(conn, r, permission))
+      .then(() => user.get(conn, 'admin'))
+      .then(res => {
+        expect(res.status).toBe(200);
+        expect(res.body).toEqual({
+          enabled: expect.any(Boolean),
```
arb: We should know this to be `true` or `false`. Same as `superuser`.

test/getUser.spec.js
```diff
@@ -0,0 +1,40 @@
...
+
+const { user } = require('../lib');
+const { generateRandomString, ConnectionFactory } = require('./setup-database');
+
+const { role } = user;
+
+describe('user.get', () => {
+  let conn;
+
+  beforeEach(() => {
```
arb: Let's stop this pattern. Just do `const conn = ConnectionFactory()` we don't do anything to mutate `conn` that would require a new one each test.

PR #106, opened 2017-08-02
🔨 - Moving tx related functions and tests

snowell:


lib/query/main.js
```diff
@@ -33,6 +33,20 @@ const execute = (conn, database, query, params) => {
...
 };
 
+const executeInTransaction = (conn, database, transactionId, query, params) => {
+  const headers = conn.headers();
+  headers.append('Accept', mimeType(query));
+  headers.append('Content-Type', 'application/x-www-form-urlencoded');
+
+  const body = buildQuery(query, params);
+
+  return fetch(conn.uri(database, transactionId, 'query'), {
```
arb: You should attach the transaction back somehow. That's what I've been doing so you can string them together. Any call that requires a transaction ID, it should come back in the result.

lib/db/main.js
```diff
@@ -119,6 +146,40 @@ const clear = (conn, database, transactionId, params = {}) => {
     .then(res => Object.assign({}, res, { transactionId }));
 };
 
+// options - graphUri, contentType, encoding
+// contentType - application/x-turtle, text/turtle, application/rdf+xml, text/plain, application/x-trig, text/x-nquads, application/trix
+// encoding - gzip, compress, deflate, identity, br
+const add = (conn, database, transactionId, content, options, params) =>
```
arb: need `options = {}`

lib/db/main.js
```diff
@@ -119,6 +146,40 @@ const clear = (conn, database, transactionId, params = {}) => {
...
+      transactionId,
+      content,
+    },
+    params
+  );
+
+// options - graphUri, contentType, encoding
+// contentType - application/x-turtle, text/turtle, application/rdf+xml, text/plain, application/x-trig, text/x-nquads, application/trix
+// encoding - gzip, compress, deflate, identity, br
+const remove = (conn, database, transactionId, content, options, params) =>
```
arb: `options = {}`

snowell:
Not sure what method you mean. I am attaching it on query.executeInTransaction...

PR #107, opened 2017-08-02
user isValid

tysoeh:
adds `user.valid` which returns 200 if creds are valid, else 401

### DIFF
--- a/lib/user/main.js
+++ b/lib/user/main.js
@@ -48,6 +48,14 @@ const changePassword = (conn, username, password, params) => {
   }).then(httpMessage);
 };
 
+const valid = (conn, params) => {
+  const headers = conn.headers();
+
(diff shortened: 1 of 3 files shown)


lib/user/main.js
```diff
@@ -41,6 +41,15 @@ const changePassword = (conn, username, password, params) => {
   }).then(httpMessage);
 };
 
+const valid = (conn, params) => {
+  const headers = conn.headers();
+  headers.append('Accept', 'application/json');
```
snowell: Unneeded. This call has an empty return body

arb: That does it return? Just status codes?

snowell: Yes, either a 200 or 401 with a Content-Length: 0 response

PR #111, opened 2017-08-07
⚡️ - Expose mechanism to override accept header

### LINKED ISSUE #109: Allow to customize accept header
Expose a mechanism for specifying the accept header yourself.

### LINKED ISSUE #110: Query string parameters not sending upstream

arb:
Closes #110
Closes #109

Also fixes bug where query string values were not properly being sent to the
server.

### DIFF
--- a/test/query.spec.js
+++ b/test/query.spec.js
@@ -19,7 +19,7 @@ describe('query.execute()', () => {
   afterAll(dropDatabase(database));
 
   it('A query result should not be empty', () =>
-    execute('select distinct ?s where { ?s ?p ?o }', {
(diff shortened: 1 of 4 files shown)


test/query.spec.js
```diff
@@ -19,7 +19,7 @@ describe('query.execute()', () => {
   afterAll(dropDatabase(database));
 
   it('A query result should not be empty', () =>
-    execute('select distinct ?s where { ?s ?p ?o }', {
+    execute('select distinct ?s where { ?s ?p ?o }', undefined, {
```
snowell: I get why you have to do this. But it's gross looking

PR #129, opened 2017-12-11
Add Virtual Graph API. Fix #125

### LINKED ISSUE #125: Virtual Graph API

snowell:
Most tests are skipped because I am unsure how to change Docker such that there's a mysql running on it, as well as a mysql JDBC jar in #STARDOG_EXT. But they do work running locally, fwiw

### DIFF
--- a/test/virtualGraphs.spec.js
+++ b/test/virtualGraphs.spec.js
@@ -1,0 +1,195 @@
+/* eslint-env jest */
+
+const vGraphs = require('../lib/virtualGraphs');
+const { ConnectionFactory } = require('./setup-database');
+
+describe('virtual_graphs', () => {
+  let conn;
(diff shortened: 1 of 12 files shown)


test/virtualGraphs.spec.js
```diff
@@ -0,0 +1,195 @@
...
+      const exists =
+        res.body.virtual_graphs &&
+        res.body.virtual_graphs.includes(`virtual://${aVGName}`);
+      if (exists) {
+        return vGraphs.remove(conn, aVGName);
+      }
+      return Promise.resolve({ status: 204 });
+    });
+
+  describe.only('list', () => {
```
tysoeh: Are we skipping all other tests because there's no relational db in Circle?

PR #146, opened 2018-02-13
Support fetch from behind proxies and/or to endpoints with self-signed certs

jmrog:
When a user is behind a proxy that has a self-signed certificate, or is directly hitting an HTTPS endpoint that has a self-signed certificate, `node-fetch` (used by our `fetch-ponyfill` dependency in Node-like environments) will, by default, reject the request with a "self-signed certificate" error. This prevents using stardog.js in Node/Electron under some fairly common development scenarios. The way to work around this with `node-fetch` is to supply an 'agent' for requests that allows connections with self-signed certificates (see [here](https://github.com/bitinn/node-fetch/issues/19), for example). However, `fetch-ponyfill` doesn't allow you to supply an agent. So, this PR puts a thin wrapper around the ponyfill, and allows end users to supply a custom agent to stardog.js when a connection is created (via the `fetchMetadata` parameter). Optionally, the end user can also supply a test function that tells stardog.js when to use the agent (the default test ensures that the agent is used only in Node-like environments, so that there aren't any issues when stardog.js is used in the browser). All of this is backwards-compatible and done in a way that requires as few changes as possible; if no agent is supplied, the original, unwrapped `fetch-ponyfill` is used as usual.

### DIFF
--- a/lib/fetch.js
+++ b/lib/fetch.js
@@ -1,47 +1,8 @@
 /* eslint-disable global-require */
-const ponyfill = require('fetch-ponyfill')();
-
-const { fetch: originalFetch } = ponyfill;
+module.exports = require('fetch-ponyfill')();
 
 // Default test for using the fetch agent.
-const isNodeLike = () =>
-  Boolean(
-    global && process && /\[native code\]/.test(process.constructor.toString())
-  );
-
-const wrapFetch = fetchMetadata => {
-  ponyfill.fetch = (url, options) => {
-    const shouldUseAgent = fetchMetadata.shouldUseAgent
-      ? fetchMetadata.shouldUseAgent(url, options)
-      : isNodeLike();
-
-    if (!shouldUseAgent) {
-      return originalFetch(url, options);
-    }
-
-    // If the user passes `agent` options for a request, just use those;
-    // otherwise, use the agent on fetchMetadata
-    const adjustedOptions = options.agent
-      ? options
-      : Object.assign({}, options, {
-          agent: fetchMetadata.agent,
-        });
-
-    return originalFetch(url, adjustedOptions);
-  };
-};
-
-const unwrapFetch = () => {
-  ponyfill.fetch = originalFetch;
-};
-
-module.exports = Object.assign({}, ponyfill, {
-  // Add an extra level of indirection so that the wrapping methods above can
-  // be called (and their effects can be seen) after this module has already
-  // been imported by other modules.
-  fetch(url, options) {
-    return ponyfill.fetch(url, options);
-  },
-});
-module.exports.wrapFetch = wrapFetch;
-module.exports.unwrapFetch = unwrapFetch;
+// const isNodeLike = () =>
+//   Boolean(
+//     global && process && /\[native code\]/.test(process.constructor.toString())
+//   );
--- a/lib/fetch.js
+++ b/lib/fetch.js
@@ -1,2 +1,47 @@
 /* eslint-disable global-require */
-module.exports = require('fetch-ponyfill')();
+const ponyfill = require('fetch-ponyfill')();
+
+const { fetch: originalFetch } = ponyfill;
+
+// Default test for using the fetch agent.
+const isNodeLike = () =>
+  Boolean(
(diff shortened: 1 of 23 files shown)


jmrog:
I should probably add tests for this.

jmrog:
Hold off on merging this, please. The more I think about it, the more I dislike it. It works as a quick fix for something in Studio, but maybe I'll just pull directly from this branch in Studio for now, until we have a better long-term fix.

The main reason I dislike this is that whether or not `fetch` is wrapped at any particular point in time is determined entirely by whatever options were passed to the _most recent_ call to `new Connection()` or `connection.config()`, and _not_ by whatever options were passed when the particular Connection object being passed to any given stardog.js API method was constructed. In other words, someone might construct a Connection object _without_ `fetchMetadata`, store a reference to it, create a new Connection object _with_ `fetchMetadata`, and then call one of the API methods with the reference to the first Connection object, and `fetch` will be wrapped for that call despite the fact that the Connection object passed in was created without `fetchMetadata`. That's not good. Unfortunately, making things happen the way they ought to happen will require a much larger number of (albeit mostly trivial!) changes to stardog.js, so that's likely to be on the back burner for a while.

jmrog:
This can be reviewed now (and merged if all's well). I went ahead and did the cleaner version of what I mentioned in my previous comment, because it wasn't as tedious as I thought it might be (only _sort of_ tedious). 😄

lib/fetch.js
```diff
@@ -1,2 +1,8 @@
 /* eslint-disable global-require */
 module.exports = require('fetch-ponyfill')();
+
+// Default test for using the fetch agent.
+// const isNodeLike = () =>
+//   Boolean(
+//     global && process && /\[native code\]/.test(process.constructor.toString())
+//   );
```
snowell: This doesn't appear to be adding much

lib/index.d.ts
```diff
@@ -39,14 +40,40 @@ declare namespace Stardog {
...
     export const version: string;
 
     /** Describes the connection to a running Stardog server. */
     export class Connection {
-        constructor(options: ConnectionOptions);
+        constructor(options: ConnectionOptions, meta?: ConnectionMeta);
 
-        config(options: ConnectionOptions): void;
+        config(options: ConnectionOptions, meta?: ConnectionMeta): void;
         headers(): Headers;
```
snowell: If we're removing the import of Headers, where does this come from? Or was it a duplicate import?

lib/index.d.ts
```diff
@@ -39,14 +40,40 @@ declare namespace Stardog {
...
+    interface ConnectionMeta {
+        /**
+         * If defined, this method is called whenever a `fetch` call is about
+         * to be made for a given Connection. The method should return a
+         * `Request` object (either a browser `Request` or a `node-fetch`
+         * `Request`) or a string URI, which will be passed as the first
+         * argument to the imminent `fetch` call. This is useful in cases where
+         * fetch needs to allow connections with self-signed certificates, for
+         * example.
+         *
```
snowell: I'm sure this is very descriptive, but I still found it a little confusing. Could be my lack of JS-fu. It also might not matter since it's all optional

lib/index.d.ts
jmrog: The previous import of `Headers` was actually wrong/broken, because it looked like this:

```typescript
import Headers from 'fetch-ponyfill';
```

That doesn't work, because `fetch-ponyfill`'s default export isn't a `Headers` instance at all (it's a function). Unfortunately, `fetch-ponyfill` doesn't export a `Headers` type. In my changed version, at least, the `Headers` type is the usual `Headers` type provided by TypeScript's `lib.es6.d.ts`. Since `fetch-ponyfill` is a ponyfill, its Headers type ought to be structurally the same, and since TypeScript uses duck typing, this should work fine (at least in most cases, if not all).

lib/index.d.ts
jmrog: I tried updating this a bit to make it clearer.

lib/fetch.js
jmrog: Whoops. Removed.

PR #181, opened 2018-11-20
Feature/180 request untransformed mappings

### LINKED ISSUE #180: Update mappings call to support requesting untransformed mappings
A feature is being added in Stardog 6.0.1 to support retrieving the original mapping string submitted to Stardog prior to any transformation. Stardog.js needs to support this.

jmrog:
Close #180.

### DIFF
--- a/lib/virtualGraphs.js
+++ b/lib/virtualGraphs.js
@@ -2,14 +2,14 @@ const { fetch } = require('./fetch');
 
 const { httpBody } = require('./response-transforms');
 
-const list = (conn, params = {}) => {
+const list = conn => {
   const headers = conn.headers();
   return fetch(conn.request('admin', 'virtual_graphs'), {
     headers,
   }).then(httpBody);
 };
 
-const add = (conn, name, mappings, options, params = {}) => {
+const add = (conn, name, mappings, options) => {
   const headers = conn.headers();
   headers.set('Content-Type', 'application/json');
   return fetch(conn.request('admin', 'virtual_graphs'), {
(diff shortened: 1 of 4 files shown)


lib/virtualGraphs.js
```diff
@@ -37,31 +37,45 @@ const update = (conn, name, mappings, options, params = {}) => {
...
   return fetch(conn.request('admin', 'virtual_graphs', name, 'options'), {
     headers,
   }).then(httpBody);
 };
 
-const mappings = (conn, name, params = {}) => {
+const mappings = (conn, name, requestOptions = {}) => {
   const headers = conn.headers();
+
+  if (requestOptions.preferUntransformed) {
```
snowell: The TS definition for the requestOptions object contains the key `untransformed` instead of `preferUntransformed`. Which do we want?

test/virtualGraphs.spec.js
```diff
@@ -192,4 +192,46 @@ describe('virtual_graphs', () => {
           expect(res.body).toContain('subjectMap');
         }));
   });
+
+  describe('mappings', () => {
+    it('accepts options that trigger a request for untransformed mappings', () =>
+      assureExists()
+        .then(() =>
+          vGraphs.mappings(conn, aVGName, { preferUntransformed: true })
```
snowell: Same comment re: `preferUntransformed`

lib/virtualGraphs.js
jmrog: whoops, good catch. `preferUntransformed` was the one intended, so I'll update.

jmrog:
@snowell PR updated. Thanks.

PR #182, opened 2019-03-28
Add `accept` param to sparql explain method

brianru:
Needed for https://github.com/stardog-union/stardog-studio/issues/757

Tests won't pass until #185


brianru:
@jmrog I value consistency a lot -- hopefully more people will use that method for the first time in the next year than are currently using it. Also there's no need for users to upgrade this second -- dealing with the breaking change is not immediately necessary to continue using the library with the latest version of Stardog.

But I don't really know our users here. How careful have we been with breaking changes in the past?

brianru:
BTW - tests will fail until stardog 6.1.4 is released

brianru:
@jmrog I updated these tests to do the right check before and after stardog 6.1.3 -- so the tests pass now and when 6.1.4 is released

jmrog:
@brianru Can we direct this PR at the `v2` branch instead of at `master` since it's a breaking API change and that's the branch that Studio uses anyway?

brianru:
Closing in favor of #191

PR #193, opened 2019-05-24
getAll and getAvailable db options

neptunian:
Not sure what the best way to go about handling people with older versions of Stardog.  Getting all the default database properties and values doesn't currently exist.  I could send back a hardcoded object?  Same with new method that gets all database properties VALUES.  send a hardcoded object of all possible db properties using existing optons.get.  I would have to know which properties belong to which version, though, so I'd need the version.  Suggestions?

### DIFF
--- a/README.md
+++ b/README.md
@@ -430,6 +430,16 @@ Returns [`Promise<HTTP.Body>`](#body)
 
 ## <a name="options">options</a>
 
+#### <a name="getavailable">`db.options.getAvailable(conn)`</a>
+
+Gets all available database options with their default values. 
+
+Expects the following parameters:
+
+- conn ([`Connection`](#connection))
+
+Returns [`Promise<HTTP.Body>`](#body)
+
 #### <a name="get">`db.options.get(conn, database, params)`</a>
 
 Gets set of options on a database. 
@@ -441,6 +451,18 @@ Expects the following parameters:
 - database (`string`)
 
 - params (`object`)
+
+Returns [`Promise<HTTP.Body>`](#body)
+
+#### <a name="getall">`db.options.getAll(conn, database)`</a>
+
+Gets all options on a database. 
+
+Expects the following parameters:
+
+- conn ([`Connection`](#connection))
+
+- database (`string`)
 
 Returns [`Promise<HTTP.Body>`](#body)
 
--- a/lib/db/options.js
+++ b/lib/db/options.js
@@ -6,11 +6,19 @@ const { httpBody } = require('../response-transforms');
 
 const dispatchDBOptions = (conn, config, body) => {
   config.headers.set('Content-Type', 'application/json');
-  return fetch(conn.request('admin', 'databases', config.database, 'options'), {
+
+  const requestOptions = {
     method: config.method,
     headers: config.headers,
-    body: JSON.stringify(flat(body, { safe: true })),
-  });
+  };
+
+  if (body) {
+    requestOptions.body = JSON.stringify(flat(body, { safe: true }));
+  }
+  return fetch(
+    conn.request('admin', 'databases', config.database, 'options'),
+    requestOptions
+  );
 };
 
 const get = (conn, database, params) => {
@@ -22,8 +30,28 @@ const get = (conn, database, params) => {
       database,
       method: 'PUT',
     },
-    DB_OPTIONS
+    // TODO now that `getAll` is available, remove use of DB_OPTIONS and require
+    // user to specify which options they want
+    DB_OPTIONS // the default list of options to GET values for
   )
+    .then(httpBody)
+    .then(res => {
+      if (res.status === 200) {
+        return Object.assign({}, res, {
+          body: flat.unflatten(res.body),
+        });
+      }
+      return res;
+    });
+};
+
+const getAll = (conn, database) => {
+  const headers = conn.headers();
+  return dispatchDBOptions(conn, {
+    headers,
+    database,
+    method: 'GET',
+  })
     .then(httpBody)
     .then(res => {
       if (res.status === 200) {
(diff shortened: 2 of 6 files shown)


brianru:
Re: previous versions of Stardog - do we need to support them? Will Studio need that support?

test/getDBOptions.spec.js
```diff
@@ -33,4 +33,16 @@ describe('db.getOptions()', () => {
         },
       });
     }));
+
+  it('should get all the options of a DB', () =>
```
brianru: good opportunity for a snapshot test?

lib/db/options.js
```diff
@@ -48,4 +74,22 @@ const set = (conn, database, databaseOptions, params) => {
   ).then(httpBody);
 };
 
-module.exports = { get, set };
+const getDbConfigProperties = (conn, params) => {
```
brianru: shorter more consistent name? `getAvailable`?

neptunian:
> Re: previous versions of Stardog - do we need to support them? Will Studio need that support?

I assumed we wanted the stardog.js api to be available to all consumers no matter the version of Stardog.  Studio will be using these new endpoints to replace the old way of doing it described below.

Currently we use this hardcoded object in Studio when we create a database: https://github.com/stardog-union/stardog-studio/blob/develop/src/common/utils/databases/dbOpts.ts
Using the endpoint would give us the properties available for that version of Stardog, otherwise we need to keep this list around for older versions.  getDbConfigProperties() is meant to replace this.

getAll() is meant to replace using this other hardcoded object
https://github.com/stardog-union/stardog.js/blob/master/lib/db/dbopts.js and using a PUT to get all possible properties with their values. 

In order to support the older versions I am thinking we just need to keep these hardcoded objects around.  In stardog.js we could check for a version and if its before the new endpoints exist we just have to do it the old way.

lib/db/options.js
neptunian: how about getMetadata?  It's different from get() and getAll() where it returns all properties with default values plus labels and description.

brianru:
I think stardog.js should provide information about available db properties for stardog versions preceding the new API method *IFF* that information is version-specific and complete. I don't think we should go back version by version to determine the available properties, so, I don't think we should do this at all in stardog.js. Studio will have to make the same assumptions it has always made about available properties -- by providing a drastically incomplete and probably for some versions inaccurate list of editable properties.

brianru:
@jmrog I took this one over, so reaching out for an independent review. The approach I ended up taking with supporting old Stardog versions is I think stardog.js should just not support the new feature for old Stardog versions. Studio will maintain a hard-coded bit of data for handling this.

README.md
```diff
@@ -460,6 +472,16 @@ Expects the following parameters:
 
 Returns [`Promise<HTTP.Body>`](#body)
 
+#### <a name="getDbConfigProperties">`db.options.getDbConfigProperties(conn)`</a>
```
anneeb: This is kind of verbose and redundant. Can we rename it to something like `db.options.getConfig` or `db.options.getConfigProperties`?

test/getDBOptions.spec.js
```diff
@@ -8,7 +8,7 @@ const {
   ConnectionFactory,
 } = require('./setup-database');
 
-describe('db.getOptions()', () => {
+describe('options.get()', () => {
```
anneeb: We should probably add a test for getConfig

README.md
brianru: agreed. what do you think about `getAvailable`?

anneeb: I'd lean towards it being somewhat analogous to the method syntax within stardog itself, but that also makes sense, or even something like `getDefaults` which is a little bit more descriptive and explicit in that includes the default values

test/getDBOptions.spec.js
brianru: done

README.md
```diff
@@ -460,6 +472,16 @@ Expects the following parameters:
 
 Returns [`Promise<HTTP.Body>`](#body)
 
+#### <a name="getAvailable">`db.options.getAvailable(conn)`</a>
+
+Gets all possible database options with their default values.
```
anneeb: There appears to be mismatch in the wording between the description text here and what's declared in the index.d.ts file. These should be generated by running `yarn docs`.

brianru:
thanks @anneeb ! I updated the docs. Do you know how to update the test image?

anneeb:
@brianru I believe it should just be updating the version for `FROM stardog-eps-docker.jfrog.io/stardog:6.1.2` in the Dockerfile

anneeb:
Also should this be pointing to the `v2` branch instead of master?

brianru:
> Also should this be pointing to the `v2` branch instead of master?

Why? Don't we generally want commits in both?

anneeb:
@brianru I know we're locked to v2 in studio, and v2 is 23 commits ahead of master. I'm actually not sure why we split it, but I believe it has something to do with release or tagging purposes. @jmrog, what's the desired workflow for this?

anneeb:
@brianru for the Dockerfile, is there a specific reason for using v6.2.3 rather than `stardog-eps-docker.jfrog.io/stardog:latest`?

Also I talked to @jmrog and it looks like we develop of the v2 branch in order to batch releases, so you'll want to rebase off of that branch and change the base of this PR to such.

brianru:
> @brianru for the Dockerfile, is there a specific reason for using v6.2.3 rather than `stardog-eps-docker.jfrog.io/stardog:latest`?

I was just sticking with the static version given that's what was there. If we change to `latest` then we will find our tests failing due to stardog breaking changes without any change to stardog.js, which is good and bad. `latest` means there's one less thing to think about, so I like it. What do you think @jmrog ?

> Also I talked to @jmrog and it looks like we develop of the v2 branch in order to batch releases, so you'll want to rebase off of that branch and change the base of this PR to such.

Huh. I thought there were actual code differences between the two, something about an http lib dependency. Previously I've been working off develop then me or @jmrog cherry-picked onto `v2`. Anyway, I'll switch it to `v2`. It's what I do locally anyway.

brianru:
Closing in favor of #198

PR #196, opened 2019-07-22
add support for `reasoning` and `description` attributes to stored queries

brianru:
See the issue for context. This is required for https://github.com/stardog-union/stardog-studio/issues/926

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -755,7 +755,9 @@ declare namespace Stardog {
             query: string,
             /** Defaults to false. */
             shared: boolean,
+            /** Attributes introduced in Stardog version 6.2.2, which are optional in versions >= 6.2.2 and ignored in earlier versions. */
             reasoning?: boolean,
+            description?: boolean,
         }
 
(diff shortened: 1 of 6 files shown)


lib/index.d.ts
```diff
@@ -754,7 +754,8 @@ declare namespace Stardog {
             database: string,
             query: string,
             /** Defaults to false. */
-            shared: boolean
+            shared: boolean,
+            reasoning?: boolean,
```
anneeb: Can you add the `description` type here too with a comment explaining why it's optional? It would also help as separation to establish that `reasoning` does not also fall into the bucket of `Defaults to false.`

Additionally, you'll want to add these to the body description for doc generation [here](https://github.com/stardog-union/stardog.js/pull/196/files#diff-83b6ae62f66182e10e7cca8617f11ac0R10) and run `yarn run docs` to update the README too.

anneeb:
@brianru this needs docs and then I think it's good to go

brianru:
@anneeb updated docs

PR #201, opened 2019-09-27
Add ability to import namespaces via string or file (new Stardog endpoints)

jmrog:
I'm not sold on calling the method for importing namespaces "set", since it doesn't _completely replace_ existing namespaces, but it was nicer in terms of corresponding to "get". What "set" actually does, as the test's filename indicates, is _add_ or _overwrite existing_ namespaces by importing from an RDF file; it doesn't _set_ the namespaces to _exactly_ those specified (e.g., if there is an existing namespace "foo=whatever" in the DB on the server, and the contents sent over HTTP consist in namespaces "bar=baz" and "blah=blah", the namespaces in the DB on the server when all is said and done will be "foo=whatever", "bar=baz", and "blah=blah" -- "foo=whatever" is not removed).

Seems okay to call it "set", though.

### DIFF
--- a/lib/db/namespaces.js
+++ b/lib/db/namespaces.js
@@ -31,13 +31,15 @@ const get = (conn, database, params) => {
       return backwardsCompatGetNamespaces(conn, database, params);
     }
 
-    return httpBody(res).then(bodyRes => ({
-      ...bodyRes,
-      body: bodyRes.body.namespaces.reduce(
-        (memo, { prefix, name }) => Object.assign({}, memo, { [prefix]: name }),
-        {}
-      ),
-    }));
+    return httpBody(res).then(bodyRes =>
+      Object.assign({}, bodyRes, {
+        body: bodyRes.body.namespaces.reduce(
+          (memo, { prefix, name }) =>
+            Object.assign({}, memo, { [prefix]: name }),
+          {}
+        ),
+      })
+    );
   });
 };
 
--- a/lib/db/namespaces.js
+++ b/lib/db/namespaces.js
@@ -1,0 +1,74 @@
+const { fetch } = require('../fetch');
+const FormData = require('form-data');
+const { get: getOptions } = require('./options');
+const lodashGet = require('lodash/get');
+
+const { httpBody } = require('../response-transforms');
+
+// For Stardog <= 7.0.2 compatibility. Remove later.
+const backwardsCompatGetNamespaces = (conn, database, params) =>
+  getOptions(conn, database, params).then(res => {
+    if (res.status === 200) {
+      const n = lodashGet(res, 'body.database.namespaces', []);
+      const names = n.reduce((memo, keyValString) => {
+        const [key, value] = keyValString.split('=');
+        return Object.assign({}, memo, { [key]: value });
+      }, {});
+      res.body = names;
+    }
+    return res;
+  });
+
+const get = (conn, database, params) => {
+  const headers = conn.headers();
+  headers.set('Accept', 'application/json');
+  return fetch(conn.request(database, 'namespaces'), {
+    method: 'GET',
+    headers,
+  }).then(res => {
+    // Old version of Stardog:
+    if (res.status === 404) {
+      return backwardsCompatGetNamespaces(conn, database, params);
+    }
+
+    return httpBody(res).then(bodyRes => ({
+      ...bodyRes,
+      body: bodyRes.body.namespaces.reduce(
+        (memo, { prefix, name }) => Object.assign({}, memo, { [prefix]: name }),
+        {}
+      ),
+    }));
+  });
+};
+
+const set = (conn, database, fileOrContents, options = {}) => {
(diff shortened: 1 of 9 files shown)


jmrog:
@brianru The Stardog changes for this haven't been released yet, so the PR is a bit premature (this is part of the reason the tests are failing on CircleCI). Feel free to run against `develop` in Stardog, or just wait until the Stardog release.

brianru:
@jmrog no worries, i run stardog `develop` locally

jmrog:
> @jmrog no worries, i run stardog `develop` locally

Okay. There are still other failures where the tests haven't been updated for Stardog 7+, but the important thing is that the namespaces tests don't fail. (The other tests are updated in the `v2-typescript` branch.)

lib/index.d.ts
```diff
@@ -669,6 +660,26 @@ declare namespace Stardog {
...
+           * @param {object} params additional parameters if needed
+           */
+          function get(conn: Connection, database: string, params?: object): Promise<HTTP.Body>;
+
+          /**
+           * Gets a mapping of the namespaces used in a database.
+           *
+           * @param {Connection} conn the Stardog server connection
+           * @param {string} database the name of the database
+           * @param {object} params additional parameters if needed
```
brianru: missing fileOrContents

lib/index.d.ts
```diff
@@ -669,6 +660,26 @@ declare namespace Stardog {
...
+           */
+          function set(conn: Connection, database: string, fileOrContents: object | string): Promise<HTTP.Body>;
```
brianru: missing `options`

lib/db/namespaces.js
```diff
@@ -0,0 +1,79 @@
...
+          (memo, { prefix, name }) =>
+            Object.assign({}, memo, { [prefix]: name }),
+          {}
+        ),
+      })
+    );
+  });
+};
+
+const set = (conn, database, fileOrContents, options = {}) => {
```
brianru: i lean towards `update` if it's not replacing, but also don't mind `set`

lib/index.d.ts
```diff
@@ -669,6 +660,26 @@ declare namespace Stardog {
...
+           * Gets a mapping of the namespaces used in a database.
+           *
+           * @param {Connection} conn the Stardog server connection
+           * @param {string} database the name of the database
+           * @param {object} params additional parameters if needed
+           */
+          function get(conn: Connection, database: string, params?: object): Promise<HTTP.Body>;
+
+          /**
+           * Gets a mapping of the namespaces used in a database.
```
brianru: better docs

lib/db/namespaces.js
brianru: as long as the docs are clear about this, it doesn't matter much to me

lib/db/namespaces.js
```diff
@@ -0,0 +1,79 @@
+const { fetch } = require('../fetch');
+const FormData = require('form-data');
+const { get: getOptions } = require('./options');
+const lodashGet = require('lodash/get');
+
+const { httpBody } = require('../response-transforms');
+
+// For Stardog <= 7.0.2 compatibility. Remove later.
+const backwardsCompatGetNamespaces = (conn, database, params) =>
+  getOptions(conn, database, params).then(res => {
```
brianru: should this only request `database.namespaces`? where you pass in an input json object with only that key.

lib/db/namespaces.js
jmrog: I opted to change it to `add`.

lib/db/namespaces.js
jmrog: It can't do that right now, at least unless I rewrite (or don't use) `options.get`, because `options.get` always sends the full `DB_OPTIONS`. Since this was the default behavior anyway, and this is only here for backwards compatibility, and changing `options.get` seems orthogonal to this PR's goal, I'd like to leave it as-is for now.

brianru: fair

lib/index.d.ts
```diff
@@ -669,6 +660,30 @@ declare namespace Stardog {
...
+           *
+           * @param {Connection} conn the Stardog server connection
+           * @param {string} database the name of the database
+           * @param {object} params additional parameters if needed
+           */
+          function get(conn: Connection, database: string, params?: object): Promise<HTTP.Body>;
+
+          /**
+           * Extracts namespaces from an RDF file or RDF string and adds any
+           * new namespaces to those that already exist in the database.
```
brianru: should you clarify that existing prefixes are updated if the namespace changed?

brianru: and I forget - is there a doc file that gets generated with this?

brianru: https://github.com/stardog-union/stardog.js/blob/master/package.json#L115

jmrog: Sure, changed.

PR #203, opened 2019-11-21
Allow `onResponseStart` to abort further stardog.js processing

jmrog:
By allowing `onResponseStart` to optionally return `false` (telling stardog.js that it should not continue processing the response itself), all processing of the Stardog response can be effectively "turned over" to the caller. This allows, for example, the caller to _stream_ the response without ever having `response.json()` (or some other response accumulator that stardog.js might call) being called on it by stardog.js.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,14 +1,9 @@
 {
-  "extends": [
-    "airbnb-base",
-    "prettier"
-  ],
+  "extends": ["airbnb-base", "prettier"],
   "parserOptions": {
     "ecmaVersion": 6
   },
-  "plugins": [
-    "prettier"
-  ],
+  "plugins": ["prettier"],
   "rules": {
     "no-unused-vars": [
       "error",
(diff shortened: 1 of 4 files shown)


brianru:
Is it right that this addresses streaming sparql queries to file, but does not touch graphql queries or exporting databases?

jmrog:
> Is it right that this addresses streaming sparql queries to file, but does not touch graphql queries or exporting databases?

Yes, that _is_ correct. We could do something similar for each, though the "hackiness" gets a bit larger each time we do. I forgot that we need to stream databases when we export, so without adding it there, some issues on the Studio side remain blocked. For that reason, I guess we should add it there at least for now.

brianru:
> > Is it right that this addresses streaming sparql queries to file, but does not touch graphql queries or exporting databases?
> 
> Yes, that _is_ correct. We could do something similar for each, though the "hackiness" gets a bit larger each time we do. I forgot that we need to stream databases when we export, so without adding it there, some issues on the Studio side remain blocked. For that reason, I guess we should add it there at least for now.

Ok. I mean, it's not 100% necessary for you to do those in this PR, as long as the door is open for whoever is taking the Studio task to implement a similar solution.

jmrog:
> Ok. I mean, it's not 100% necessary for you to do those in this PR, as long as the door is open for whoever is taking the Studio task to implement a similar solution.
Yeah, sounds right to me. Perhaps I'll leave as-is for the moment, then, for the sake of more "atomic" PRs. But yes, the door is open to do the same thing that is done here in either of the other cases.

jmrog:
@brianru This is ready now. The test failures are the expected ones when we're hitting Stardog 7+ (ICV, versioning, DB copying, and some GraphQL identifier changes). (Those test failures are resolved in my `v2-typescript` branch so that'll be taken care of when we finally merge.)

PR #212, opened 2020-01-27
Add support for streaming export data to file (and compression)

### LINKED ISSUE #210: Add support for streaming "export data" directly to disk
This should work just like it does for streaming query results directly to disk -- i.e., accept an additionalHandler that can tell stardog.js not to do any further processing and just return the raw streaming response.

jmrog:
Close #210 .

Test failures are the expected failures for Stardog 7.x and stardog.js pre-v2 release.

Re-upping PR against the correct branch!

### DIFF
--- a/lib/db/main.js
+++ b/lib/db/main.js
@@ -203,18 +203,42 @@ const remove = (
 //  mimeType
 // params
 //  graphUri
-const exportData = (conn, database, options = {}, params = {}) => {
+const exportData = (
+  conn,
+  database,
+  options = {},
+  params = {},
+  additionalHandlers = {}
+) => {
   const headers = conn.headers();
   headers.set('Accept', options.mimetype || 'application/ld+json');
 
   const queryParams = {
     'graph-uri': params.graphUri || 'tag:stardog:api:context:all',
   };
+
+  if (params.compression) {
+    queryParams['server-side'] = true; // required for using compression
+    queryParams.compression = params.compression;
+  }
+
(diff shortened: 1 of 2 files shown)


lib/db/main.js
```diff
@@ -203,18 +203,42 @@ const remove = (
...
-  return fetch(conn.request(database, suffix), {
+  const fetchPromise = fetch(conn.request(database, suffix), {
     headers,
-  }).then(httpBody);
+  });
+
+  return fetchPromise.then(res => {
+    const shouldContinue = !additionalHandlers.onResponseStart
+      ? true
+      : additionalHandlers.onResponseStart(res);
```
anneeb: Do you think we'll ever need to handle a case where `onResponseStart` is asynchronous?

Also, not a big deal, but I find `const x = condition ? ifTrue : ifFalse` to be more readable than `const x = !condition ? ifFalse : ifTrue`

jmrog: Maybe (in both cases), but I just used the same API/code as is currently used in `lib/query/main.js`. Don't really want to change in either case unless you'd prefer I also make changes there. Doesn't seem too pressing to me, since this code is basically "stop-gap" code until we get stardog.js 2.0.

anneeb: Sure, I'm fine with that.

jmrog: https://github.com/stardog-union/stardog.js/blob/v2/lib/query/main.js#L32 is the code I meant, incidentally.

PR #221, opened 2020-03-19
Add database constraint when adding/updating vg

### LINKED ISSUE #220: Add support for creating/updating virtual graphs with database id
Backend already supports this, just need to add 'db' to the body of the requests.

joshhk72:
Close #220

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -1110,13 +1110,13 @@ declare namespace Stardog {
         type AllVgOptions = SharedOptions & RdbmsOptions & MongoOptions & CsvOptions;
         
         interface MappingsRequestOptions {
-            preferUntransformed?: boolean;
-            syntax?: string;
+          preferUntransformed?: boolean;
+          syntax?: string;
         }
         
         interface VgMeta { 
-            db?: string; 
-        };
+          db?: string; 
+        }
 
         /**
          * Retrieve a list of virtual graphs
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -1134,7 +1134,7 @@ declare namespace Stardog {
          * @param {Options} options the JDBC (and other) options for the graph
          * @param {VgMeta} meta database name and other graph specifications
          */
-        function add<T extends AllVgOptions>(conn: Connection, name: string, mappings: string, options: T, meta?: Meta): Promise<HTTP.Body>;
+        function add<T extends AllVgOptions>(conn: Connection, name: string, mappings: string, options: T, meta?: VgMeta): Promise<HTTP.Body>;
 
         /**
          * Update a virtual graph in the system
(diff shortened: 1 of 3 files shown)


lib/virtualGraphs.js
```diff
@@ -9,7 +9,7 @@ const list = conn => {
   }).then(httpBody);
 };
 
-const add = (conn, name, mappings, options) => {
+const add = (conn, name, mappings, options, params) => {
   const headers = conn.headers();
```
joshhk72: A question would be whether to pass in 'db' through params, or whether it should be passed in by itself.

lib/virtualGraphs.js
```diff
@@ -18,12 +18,13 @@ const add = (conn, name, mappings, options) => {
       name,
       mappings,
       options,
+      db: params.db || null,
```
joshhk72: I'm not sure if the null is possible, perhaps both 'null' and 'undefined' are converted to 'null' in Java?

jmrog: `JSON.stringify` will actually leave out `undefined` values entirely. So, I think the null is unnecessary (same for the instance below this one).

lib/virtualGraphs.js
jmrog: So, my thoughts here are these:

- "params" is something we use typically (if not exclusively) for query parameters. Since this isn't a query parameter, I don't think it should be on "params".
- However, I also don't think it should be on "db" (or equivalent) by itself, because that isn't very forward-thinking (we might want to add additional options in the future, and we don't want them to all be named arguments).
- So, how about an argument not named "params" that is still an object on which various bits of metadata can be passed in? I'd support calling it "meta" or "metadata".

lib/virtualGraphs.js
```diff
@@ -9,7 +9,8 @@ const list = conn => {
   }).then(httpBody);
 };
 
-const add = (conn, name, mappings, options) => {
+const add = (conn, name, mappings, options, meta) => {
+  const { db } = meta;
```
jmrog, line 13: This will throw if `meta` is undefined, as it will be for most existing calls to this method, and I don't think we want to make it required (right?). So, suggestion:

```suggestion
const add = (conn, name, mappings, options, meta = {}) => {
  const { db } = meta;
```

lib/virtualGraphs.js
```diff
@@ -18,12 +19,14 @@ const add = (conn, name, mappings, options) => {
...
       options,
+      db,
     }),
     headers,
   }).then(httpBody);
 };
 
-const update = (conn, name, mappings, options) => {
+const update = (conn, name, mappings, options, meta) => {
+  const { db } = meta;
```
jmrog, line 29: Same suggestion as above:

```suggestion
const update = (conn, name, mappings, options, meta = {}) => {
  const { db } = meta;
```

joshhk72:
Sorry, also realized that I should try to update docs for both readme + declarations

joshhk72:
Updating docs to be more in line with other stuff, in case someone might be looking (will delete comment when done)

joshhk72:
Another question is, is it necessary to somehow specify that "VgMeta" is optional in the readme?

lib/index.d.ts
```diff
@@ -1108,11 +1108,15 @@ declare namespace Stardog {
         }
 
         type AllVgOptions = SharedOptions & RdbmsOptions & MongoOptions & CsvOptions;
-
+
```
jmrog, line 1111: Some formatting isn't right here (extra spaces, and 4 spaces vs. 2).

joshhk72: oops, think i never changed my defaults for mac, and got into an overreliance on Prettier.

joshhk72:
!!!! I just realized that this should be branched from v2 (I remembered learning about why stardog.js has separate branches for v2 during my first week, but forgot until this point)

PR #224, opened 2020-03-23
Add database model endpoint

### LINKED ISSUE #217: Add support for Stardog data model endpoint
This is the HTTP equivalent of the Stardog CLI command `data model`.

anneeb:
close #217

### DIFF
--- a/README.md
+++ b/README.md
@@ -349,6 +349,20 @@ Expects the following parameters:
 
 Returns [`Promise<HTTP.Body>`](#body)
 
+#### <a name="model">`db.model(conn, database, params)`</a>
+
+Gets the model for a database.
+
+Expects the following parameters:
+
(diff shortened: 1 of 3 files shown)


anneeb:
> Does this actually work for you? I had work locally that I hadn't pushed yet that was gonna do this -- note the comment inline:
> 
> In other words, for me it was not working without specifying a `schema` query param. I'm on a development version of Stardog, though, so it could be they changed this.

In 7.2 at least, it appears that you either need `schema: default` or `reasoning: true` to have the request succeed. It also looks like having `reasoning.true` overwrites whatever `schema.default` does so maybe always setting a default value here would be okay? I'm not sure how important keeping this parameter blank is, and can't seem to an answer in the Stardog code.

PR #232, opened 2020-08-05
Allow queries to be stored with annotations

joshhk72:


joshhk72:
Upon further inspection, may not work with current http request(?). May have to look a bit more

Issue #136, opened 2018-01-10
Integration tests for virtualGraph methods


tysoeh:
Add a sql db to our docker image in circleci so that we can test virtualGraph methods properly.

sfkrauss:
CC @jmrog, Brian and I propose moving to stardog.js backlog

jmrog:
Sounds good.

PR #234, opened 2020-10-05
[STUD-414] Make all tests work with Stardog 7+

jmrog:


joshhk72:
Maybe super dumb question, but should the tests only be expected to pass for the latest versions of Stardog?

joshhk72:
Looks like it's mostly with the JSON-LD stuff, not quite sure what might be the best option to resolve this.

jmrog:
> Looks like it's mostly with the JSON-LD stuff, not quite sure what might be the best option to resolve this.

I think this is mostly okay (for now), but fill free to push back. So far, stardog.js has always been tested against only the latest Stardog. And the failures that you see when you run the tests against earlier versions of Stardog (e.g., 7.2.1 or 6.x.x) are not failures in functionality _per se_, but are just "failures" because the shapes of Stardog's responses have changed in some cases (I suppose it could be argued that we should therefore make the tests less dependent on the response shapes -- but I'm a little hesitant about that because they currently provide some extra sanity and have caught integration-related issues for us before). Additionally, we're going to release this as part of a major version bump in stardog.js (2.x.x), so no one should be upgrading and _expecting_ that there aren't breaking changes anyway. Combine all of those factors together and I think that just getting things back to a working state with the latest Stardog is a sufficient goal for this PR.

That said, we might want to make a separate push to test stardog.js against earlier versions of Stardog as well, and to make some kind of "contract" regarding what is and isn't supported. I think that can be a separate push, though. But again, feel free to push back. (Also tagging in @anneeb, @cchaffatt, and @lafirey here to see whether they have any thoughts about this.)

anneeb:
> I think this is mostly okay (for now), but fill free to push back. So far, stardog.js has always been tested against only the latest Stardog. And the failures that you see when you run the tests against earlier versions of Stardog (e.g., 7.2.1 or 6.x.x) are not failures in functionality _per se_, but are just "failures" because the shapes of Stardog's responses have changed in some cases (I suppose it could be argued that we should therefore make the tests less dependent on the response shapes -- but I'm a little hesitant about that because they currently provide some extra sanity and have caught integration-related issues for us before). Additionally, we're going to release this as part of a major version bump in stardog.js (2.x.x), so no one should be upgrading and _expecting_ that there aren't breaking changes anyway. Combine all of those factors together and I think that just getting things back to a working state with the latest Stardog is a sufficient goal for this PR.
> 
> That said, we might want to make a separate push to test stardog.js against earlier versions of Stardog as well, and to make some kind of "contract" regarding what is and isn't supported. I think that can be a separate push, though. But again, feel free to push back. (Also tagging in @anneeb, @cchaffatt, and @lafirey here to see whether they have any thoughts about this.)

I think for the sake of this release, only testing against latest Stardog is okay (this is consistent with the testing expectations we set for Studio currently). I agree that setting up the infrastructure to test against non-latest Stardog should be a separate push, as it will probably require some changes to our snapshots in addition to the CircleCI flows. There's definitely room to improve our documentation, and I think an obvious start would be to explicitly state the compatibility expectations between Stardog.js and Stardog versions, as well as any notes about known deprecation or otherwise breaking changes. I'd lean towards starting to introduce those documentation improvements now though, as they'd pair nicely with the major release.

joshhk72:
> explicitly state the compatibility expectations between Stardog.js and Stardog versions

Second this, if plausible.

jmrog:
Our summary of things right now exists here: https://github.com/stardog-union/stardog.js#version-details.

Seems like we are/were saying basically that stardog.js 1.x.x works with Stardog 5+, whereas earlier Stardog versions will work with stardog.js 0.x.x. I think we now want to say that stardog.js 2.x.x works with Stardog 7+, stardog.js 1.x.x works with Stardog 5 and 6 (and most of 7, but we won't officially support it there?), and anything earlier than 5 may work with stardog.js 0.x.x. We can probably put this into an easy-to-read table.

Worth noting: I believe that Stardog is officially no longer maintaining versions prior to 7.x.x after the end of this year.

lafirey:
> Worth noting: I believe that Stardog is officially no longer maintaining versions prior to 7.x.x after the end of this year.
Correct -- support and maintenance prior to 7 is ending at the end of the year. 

The above seems correct to me. As long as we are clear on what version of stardog, stardog.js 2.x.x supports that is fine. I am a little unclear on the before/after json-ld changes in 7.2.1. Would let @jmrog make the final call but I am not opposed to us specifying 7.3 and above for stardog.js 2.x.x. Or at least noting the failures before 7.3

jmrog:
@joshhk72 I updated this PR to include some of the requested documentation clarifications. Primarily, I added this:

image

Is there anything else blocking the merge here? (Note that this is the merge into `v2`, not the merge into `master`, so we still have time to let things brew there.) I'd like to get this in somewhat quickly, if possible, so that other stardog.js work based on `v2` isn't held up (or required to branch off of a branch of a branch 😄 ).

test/optimizeDB.spec.js
```diff
@@ -24,7 +24,10 @@ describe('optimizeDB()', () => {
       expect(res.status).toEqual(404);
     }));
 
-  it('should optimize an online DB', () =>
+  // TODO: This case is currently skipped because there is no way in Stardog
+  // 7.3.x to ensure that the DB is ready for optimization (so, this often
+  // errors out). The erroring out should be fixed in Stardog 7.4.1.
```
joshhk72, line 29: has this been fixed in the 7.4.1 release?

lib/db/main.js
```diff
@@ -106,6 +106,17 @@ const optimize = (conn, database, params) => {
   }).then(httpBody);
 };
 
+const copy = (conn, database, destination, params) => {
```
joshhk72: What's this for?

joshhk72 (commented):
Some things I was wondering:

1) Is there a reason we're skipping most of the vg tests?

2) Should we drop databases after the test? (may not matter since we're not expecting to run these a lot locally?)

lib/db/main.js
jmrog: Are you looking at a specific commit? This change shouldn't be in the PR (and I don't see it there). I had accidentally removed the `copy` method and then added it back.

joshhk72: Oops, just realized the last commit after

jmrog:
> Some things I was wondering:
> 
> 1. Is there a reason we're skipping most of the vg tests?

Yeah. stardog.js tests don't currently set up an actual VG for the tests to work with. There is an open issue for that here: https://github.com/stardog-union/stardog.js/issues/136

> 2. Should we drop databases after the test? (may not matter since we're not expecting to run these a lot locally?)

Which test? The test suites should have an `afterAll(dropDatabase(database))` call within them; the ones I checked quickly do have that, but I definitely could have missed one.

test/optimizeDB.spec.js
jmrog: Looks like it! Updated the PR.

joshhk72:
Ah, I think it leaves one database, but don't think it's too big of a deal for now.

joshhk72 (approved):
LGTM

PR #235, opened 2020-10-12
[STUD-490] Add cluster api support

rmyers:
Adding support for stardog cluster api calls:

* GET /admin/cluster - cluster info
* GET /admin/cluster/status - cluster status

I have added some tests that mock the cluster responses as it seems that setting up a cluster for testing will require a lot of setup. I think this can be done with a new docker image or setup the existing one to create a cluster. 

I based this off of the 'fix tests for 7' branch so probably easiest to look at the last commit for reviewing this.


rmyers:
PR on the wrong branch closing

PR #236, opened 2020-10-12
[STUD-490] add cluster api support

rmyers:
Adding support for stardog cluster api calls:

GET /admin/cluster - cluster info
GET /admin/cluster/status - cluster status

I have added some tests that mock the cluster responses as it seems that setting up a cluster for testing will require a lot of setup. I think this can be done with a new docker image or setup the existing one to create a cluster.

I based this off of the 'fix tests for 7' branch so probably easiest to look at the last commit for reviewing this.


jmrog:
@rmyers For the purposes of cleaner review, can you rebase commit `a609373` on the `v2` branch when you have a chance? The "fix tests" branch has been merged into `v2` now. Thanks!

test/clusterInfo.spec.js
```diff
@@ -0,0 +1,55 @@
...
+
+    expect(Fetch.fetch).toHaveBeenCalledTimes(1);
+    expect(Fetch.fetch).toHaveBeenCalledWith(
+      'http://localhost:5820/admin/cluster',
+      {
+        headers: expectedHeaders,
+      }
+    );
+
+    done();
```
jmrog, line 53: We try to avoid having to call `done` manually in the stardog.js tests (mostly just another convention, but it does help a little bit to avoid the "oops, I forgot" case, and it also removes the need to write your own `catch` clause). With promises, Jest can handle this itself. You could do:

```javascript
const clusterPromise = cluster.info(conn);
expect(Fetch.fetch).toHaveBeenCalledTimes(1);
/* . . . etc . . . */
return clusterPromise.then(res => { /* . . . etc . . . */ });
```

test/optimizeDB.spec.js
```diff
@@ -24,7 +24,10 @@ describe('optimizeDB()', () => {
       expect(res.status).toEqual(404);
     }));
 
-  it('should optimize an online DB', () =>
+  // TODO: This case is currently skipped because there is no way in Stardog
+  // 7.3.x to ensure that the DB is ready for optimization (so, this often
+  // errors out). The erroring out should be fixed in Stardog 7.4.1.
+  it.skip('should optimize an online DB', () =>
```
jmrog, line 30: This part will need to be removed from the PR, as it's stale.

jmrog (changes requested):
Looks good, but a few requests for changes. Most of them are inline, but here's an additional one: you'll need to add type information for the `cluster` calls to "index.d.ts".

lib/cluster.js
```diff
@@ -0,0 +1,24 @@
...
+  headers.set('Accept', 'application/json');
+  return Fetch.fetch(conn.request('admin', 'cluster', 'status'), {
+    headers,
+  }).then(httpBody);
+};
+
+module.exports = {
+  info,
+  status,
+};
```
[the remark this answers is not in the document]
rmyers: I changed the import to make it possible to mock the fetch function. Since it needs an Object to spyOn, couldn't get it to work with a direct import. But I'm working on getting the cluster to run with docker compose and then I can change the tests to a full integration instead of mocks to address this.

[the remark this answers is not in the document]
rmyers: I was able to get a simple cluster to run and have updated the tests to use the cluster nodes for the connection without mocking fetch. This should address the code style issues.

rmyers:
@jmrog I have added a sample docker-compose and addressed the pr comments. When you have time please have a look. I'm not exactly sure how to get this going in circleci as I'm guessing that would break all the pr's and branches that don't have this code. You can however run this locally to test it out:

```bash
docker-compose -f .circleci/docker-compose.yml up 
```

then in another terminal run:

```bash
npm run test:cluster
```

test/cluster/clusterInfo.spec.js
```diff
@@ -0,0 +1,34 @@
...
+const { ConnectionFactory } = require('../setup-database');
+
+describe('cluster.info()', () => {
+  it('should retrieve a JS object containing cluster information', () => {
+    const coordinatorConnection = ConnectionFactory(5821);
+
+    return cluster.info(coordinatorConnection).then(res => {
+      const { nodes } = res.body;
+      const coordinator = nodes.find(node => node.search(/5821/) > 0);
+      const participant = nodes.find(node => node.search(/5822/) > 0);
```
jmrog, line 13: Nitpick-ish: Instead of `node.search(regex) > 0`, you can just more directly do `node.includes(number)` here to accomplish the same purpose a little more clearly.

test/cluster/clusterStatus.spec.js
```diff
@@ -0,0 +1,42 @@
...
+const { ConnectionFactory } = require('../setup-database');
+
+describe('cluster.status()', () => {
+  it('should retrieve a JS object containing cluster status', () => {
+    const coordinatorConnection = ConnectionFactory(5821);
+
+    return cluster.status(coordinatorConnection).then(res => {
+      const { nodes } = res.body;
+      const coordinator = nodes.find(node => node.address.search(/5821/) > 0);
+      const participant = nodes.find(node => node.address.search(/5822/) > 0);
```
jmrog, line 13: Same comment here (and below) re: `includes`.

jmrog (changes requested):
This looks great! I made a nitpicky comment inline that I'd like to have addressed before merging, but otherwise seems ready to go. (I can handle updating CircleCI-related stuff as necessary.)

jmrog (approved):
LGTM

PR #237, opened 2020-10-27
DO NOT MERGE: v2.0.0 candidate for Stardog 6

jmrog:
This branch contains the proposed API for v2.0.0 of stardog.js (for compatibility with Stardog 6). It is intended to be essentially the current `v2` branch with methods (and options) that work only in Stardog 7+ removed. Seeking approval before cutting the release and tagging as v2.0.0. I do not think we should actually merge into `master`.

Tests should work with the latest release of Stardog 6.

Changes to these files have already been approved in discrete reviews. You can see what was _removed_ from `v2` for this release by looking at the commit history and the various "Revert" commits toward the end.


jmrog:
Note to self/others: we will need to update CHANGELOG.MD before release.

anneeb (approved):
LGTM

Dockerfile
```diff
@@ -1,3 +1,3 @@
-FROM stardog-eps-docker.jfrog.io/stardog:6.1.2
+FROM stardog-eps-docker.jfrog.io/stardog:6.2.4
```
anneeb, line 1: do we want this any higher?

jmrog: Is there a higher one for v6? This is the highest release for that major version that I saw in the release notes.

anneeb: oh yeah, maybe not. I was going off the binaries in artifactory, where I see 6.2.7

jmrog:
This was released the week of 2/8. As the title notes, we are not merging this into `master`. Therefore, closing the PR.

anneeb:
@jmrog are we missing a github tag for this? i see 2.0.0 in npm but not in  the github tags

PR #238, opened 2020-10-28
v3.0.0 candidate for Stardog 7

jmrog:
UPDATE: This now seems ready to be merged. We have been using the branch in Studio for a while now. In addition, I have:

1. Manually tested Studio against both Stardog 6 and Stardog 7 with this version of stardog.js linked.
2. Run Studio's browser integration tests (all passing) against both Stardog 6 and Stardog 7 with this version of stardog.js linked.

OLD MESSAGE: This is the current proposal for the v3 release of stardog.js. I _do_ think we should merge this one into master before releasing (which is different from what we'll do for the v2 release).

This branch is basically our current `v2` branch, with a few things that no longer work in Stardog 7 removed.


anneeb (approved):
LGTM

PR #239, opened 2020-10-28
[STUD-528] Add Data Sources

anneeb:
[STUD-528]

You can run the full test suite by removing the `describe.only` as long as you have a running MySQL server with an empty password for the root user on port 3306 (default for MySQL) and the [JDBC Driver for MySQL](https://dev.mysql.com/downloads/connector/j/) in copied into your `stardog_install_dir/server/dbms/` (see [Stardog Manual](https://www.stardog.com/docs/#_configuration)).

[STUD-528]: https://stardog.atlassian.net/browse/STUD-528

### DIFF
--- a/test/dataSources.spec.js
+++ b/test/dataSources.spec.js
@@ -32,7 +32,7 @@ describe('data_sources', () => {
       if (!exists) {
         return dataSources.add(conn, aDSName, aOptions);
       }
-      return Promise.resolve({ status: 201 });
+      return res;
     });
 
   const assureNotExists = () =>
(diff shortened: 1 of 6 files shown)


test/dataSources.spec.js
```diff
@@ -0,0 +1,154 @@
...
+    });
+
+  const assureNotExists = () =>
+    dataSources.list(conn).then(res => {
+      const exists =
+        res.body.data_sources && res.body.data_sources.includes(aDSName);
+      if (exists) {
+        return dataSources.remove(conn, aDSName);
+      }
+      return Promise.resolve({ status: 204 });
```
jmrog, line 45: A little nitpicky, but for these two methods, why not just `return res` at the end? That way we don't have to put an additional Promise in the flow here, and we don't have to ensure that our hard-coded numbers always match the API's status codes.

anneeb: ah, sure

jmrog (approved):
LGTM

PR #240, opened 2020-11-06
Add VG online + data source metadata

anneeb:


lib/virtualGraphs.js
```diff
@@ -18,33 +18,41 @@ const listInfo = conn => {
...
+  const { db, dataSource } = meta;
+  const body = {
+    name,
+    mappings,
+    options,
+    db,
+  };
+  if (dataSource) {
+    body.data_source = dataSource;
+  }
```
jmrog, line 30: It looks like we always send `db` even though it isn't require/could be `undefined`. But we don't do the same here for `data_source`. Is that because the API doesn't handle it correctly? Or is there some other reason for the different behavior?

anneeb: hm, it does appear that sending a data source is always required in the newest API, and providing the option doesn't seem to break older versions. i'll update it to match the db.

jmrog:
@anneeb This looks good to me, but can you fix the build failure?

anneeb:
> @anneeb This looks good to me, but can you fix the build failure?

oo, yep

jmrog (approved):
:shipit:

PR #241, opened 2020-11-06
[STUD-517] data hub - allow stardog.js to accept query params

cchaffatt:
close [STUD-517](https://stardog.atlassian.net/browse/STUD-517?atlOrigin=eyJpIjoiYWIwMWFlZGM1ZWQ1NDBlOTkwMWFiOThmNjMxNzI0YWYiLCJwIjoiaiJ9)

### DIFF
--- a/lib/dataSources.js
+++ b/lib/dataSources.js
@@ -1,4 +1,5 @@
 const { fetch } = require('./fetch');
+const qs = require('querystring');
 
 const { httpBody } = require('./response-transforms');
 
@@ -50,9 +51,12 @@ const update = (conn, name, options) => {
   }).then(httpBody);
 };
 
-const remove = (conn, name) => {
+const remove = (conn, name, params = {}) => {
   const headers = conn.headers();
-  return fetch(conn.request('admin', 'data_sources', name), {
+  const queryParams = Object.keys(params).length
+    ? `?${qs.stringify(params)}`
+    : '';
+  return fetch(conn.request('admin', 'data_sources', name, queryParams), {
     method: 'DELETE',
     headers,
   }).then(httpBody);
--- a/lib/dataSources.js
+++ b/lib/dataSources.js
@@ -56,7 +56,7 @@ const remove = (conn, name, params = {}) => {
   const queryParams = Object.keys(params).length
     ? `?${qs.stringify(params)}`
     : '';
-  return fetch(conn.request('admin', 'data_sources', name + queryParams), {
+  return fetch(conn.request('admin', 'data_sources', `${name}${queryParams}`), {
     method: 'DELETE',
     headers,
   }).then(httpBody);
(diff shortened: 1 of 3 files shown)


lib/dataSources.js
```diff
@@ -50,9 +51,12 @@ const update = (conn, name, options) => {
   }).then(httpBody);
 };
 
-const remove = (conn, name) => {
```
anneeb, line 53: can you also add the docs to the `index.d.ts` file and run `yarn docs` to update the readme?

lib/dataSources.js
```diff
@@ -50,9 +51,12 @@ const update = (conn, name, options) => {
...
+const remove = (conn, name, params = {}) => {
   const headers = conn.headers();
-  return fetch(conn.request('admin', 'data_sources', name), {
+  const queryParams = Object.keys(params).length
+    ? `?${qs.stringify(params)}`
+    : '';
+  return fetch(conn.request('admin', 'data_sources', name, queryParams), {
```
anneeb, line 59: I'm not sure about this since it's adding an extra slash between the name and query params: 
`http://localhost:5820/admin/data_sources/MyDataSource/?force=true`

cchaffatt: yes, it keeps the ending slash but does work as expected

cchaffatt: we can append to the data source name to make it look better

anneeb: i think that would be preferable (and seems to be consistent with the rest of the uris we generate)

cchaffatt: It looks like we are doing it both ways. There is a `suffix` as the last argument and some in places appending.

cchaffatt: actually it looks like those `suffix`'s have names attached.. just kidding

lib/index.d.ts
```diff
@@ -1343,6 +1343,7 @@ declare namespace Stardog {
          *
          * @param {Connection} conn the Stardog server connection
          * @param {string} name the data source name
+         * @param {object} params the query string parameters
          */
         function online(conn: Connection, name: string): Promise<HTTP.Body>;
```
anneeb, line 1348: it looks like we're still missing the params param here as well as the readme update

lib/dataSources.js
```diff
@@ -50,9 +51,12 @@ const update = (conn, name, options) => {
...
 };
 
-const remove = (conn, name) => {
+const remove = (conn, name, params = {}) => {
   const headers = conn.headers();
-  return fetch(conn.request('admin', 'data_sources', name), {
+  const queryParams = Object.keys(params).length
+    ? `?${qs.stringify(params)}`
+    : '';
+  return fetch(conn.request('admin', 'data_sources', name + queryParams), {
```
anneeb, line 59: we generally use template literals rather than addition for strings

lib/index.d.ts
cchaffatt: looks like I added it to the wrong method too.. oops

anneeb (approved):
LGTM

PR #248, opened 2020-12-03
[STUD-XXX] Add force option to dataSource.update

joshhk72:


lib/dataSources.js
```diff
@@ -38,12 +38,13 @@ const add = (conn, name, options) => {
   }).then(httpBody);
 };
 
-const update = (conn, name, options) => {
+const update = (conn, name, options, params) => {
```
anneeb, line 41: I think we usually equate `params` to be `queryParams` (`ex.com?query=query`), so this may be better named as something like `opts`. Otherwise, I do lean towards not append to the body unless a parameter is explicitly defined, but don't feel too strongly about that since I think we do that elsewhere but I can't remember where.

joshhk72: Hmm, I like `requestOptions` (currently only used in `virtualGraphs.mappings`)

anneeb: ooh yep, that's a good way to differentiate it from the actual object options, nice

lib/dataSources.js
```diff
@@ -38,15 +38,18 @@ const add = (conn, name, options) => {
...
 };
 
-const update = (conn, name, options) => {
+const update = (conn, name, options, requestOptions = {}) => {
   const headers = conn.headers();
   headers.set('Content-Type', 'application/json');
+
+  const body = { name, options };
+  if (requestOptions.force) {
+    body.force = true;
```
anneeb, line 47: i would set this to be the value of the option rather than always true when present (so this way if you have it as `requestOptions.force = false` it passes it through rather than making it true)

anneeb (approved):
LGTM

Issue #255, opened 2021-03-16
new Buffer() deprecation


jologz:
I haven't looked closely on where `add` method is being used but it contains a `new Buffer(fileContents)` found [here](https://github.com/stardog-union/stardog.js/blob/5a61680525f978fa1772b1846060590900590ada/lib/db/docs.js#L26).  It can be changed to `Buffer.from(fileContents)`

I bundled my [TS project](https://github.com/jologz/stardog-named-graph-alias) and received a warning about buffer deprecation.

```
[DEP0005] DeprecationWarning: Buffer() is deprecated due 
to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
``` 

Stardog.js showed it and some other plugins.  Its a warning so it is a low priority, IMO.

PR #256, opened 2021-03-16
Update `new Buffer` to `Buffer.from(...)`

jologz:
I haven't tested this.  I'm not sure where `add` is used.  This will resolve issue #255


jmrog (approved):
Looks good. Thanks!

PR #260, opened 2021-04-22
[VET-867] feature: add token auth support

rmyers:
* Add user.token() command to fetch a jwt token
* Add support for using token for Authorization header


jmrog (approved):
LGTM!

PR #262, opened 2021-06-15
Fix capitalization for mimetype in docs

joshhk72:


jmrog:
@joshhk72 I don't think this is right. The (lack of) capitalization seems correct, based on the implementation here: https://github.com/stardog-union/stardog.js/blob/master/lib/db/main.js#L213

joshhk72:
@jmrog this is removing the capitalization, since it's `mimeType` in the docs.

jmrog:
> @jmrog this is removing the capitalization, since it's `mimeType` in the docs.

Oh, 🤦 . I read it in reverse.

jmrog (approved):
LGTM

PR #266, opened 2022-05-20
use empty string for password if it is `undefined`

SpiralP:
otherwise we use the string `"undefined"` for the password


joshhk72 (approved):
oh........ 

LGTM

anneeb (approved):
LGTM

PR #267, opened 2022-05-20
[VET-1899] Deprecate ICV add/remove

SpiralP:
VET-1899

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -377,6 +377,9 @@ declare namespace Stardog {
             /**
              * Adds integrity constraints to a given database.
              *
+             * @deprecated Support for storing ICV constraints in the system database is deprecated in Stardog 7.5.0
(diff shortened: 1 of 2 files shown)


lib/index.d.ts
```diff
@@ -377,6 +377,9 @@ declare namespace Stardog {
             /**
              * Adds integrity constraints to a given database.
              *
+             * @deprecated Support for storing ICV constraints in the system database is deprecated in Stardog 7.5.0
+             * and removed in Stardog 8.0.0; instead, SHACL constraints can be managed using {@link db.add}.
+             *
```
SpiralP, line 382: did we want to give console.log's when you try to run these?

anneeb: i think this is fine as-is

anneeb (approved):
LGTM

alex-r-bigelow (approved):

PR #276, opened 2022-08-05
[VET-2029] Stardog.js changes for automapping service

SpiralP:
VET-2029

* requires develop version of stardog (probably future v8.1.0)

### DIFF
--- a/lib/dataSources.js
+++ b/lib/dataSources.js
@@ -167,6 +167,18 @@ const refreshMetadata = (conn, name, tableName = '') => {
   ).then(httpBody);
 };
 
+const suggestions = (conn, trigBody) => {
+  const headers = conn.headers();
+  headers.set('Content-Type', 'application/trig');
+  headers.set('Accept', 'application/ld+json');
+
+  return fetch(conn.request('matcher', 'suggestions'), {
+    method: 'POST',
+    body: trigBody,
+    headers,
+  }).then(httpBody);
+};
+
 module.exports = {
   available,
   query,
@@ -181,6 +193,7 @@ module.exports = {
   list,
   add,
   options,
+  suggestions,
 
   getMetadata,
   updateMetadata,
--- a/lib/dataSources.js
+++ b/lib/dataSources.js
@@ -167,14 +167,14 @@ const refreshMetadata = (conn, name, tableName = '') => {
   ).then(httpBody);
 };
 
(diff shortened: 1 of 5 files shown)


lib/dataSources.js
```diff
@@ -167,6 +167,18 @@ const refreshMetadata = (conn, name, tableName = '') => {
   ).then(httpBody);
 };
 
+const suggestions = (conn, trigBody) => {
+  const headers = conn.headers();
+  headers.set('Content-Type', 'application/trig');
+  headers.set('Accept', 'application/ld+json');
```
anneeb, line 173: i would maybe make this configurable (`(conn, config, content)`)

package.json
```diff
@@ -133,5 +133,5 @@
     "singleQuote": true,
     "trailingComma": "es5"
   },
-  "stardog-version": ">=7.0.0"
+  "stardog-version": ">=8.1.0"
```
anneeb, line 136: i think this is fine as `>=8.0.0`

lib/dataSources.js
anneeb: also, can you update the docs with this?

SpiralP: configurable how? into fetch? why?

package.json
SpiralP: you won't be able to use this new endpoint on 8.0.0 though; is that ok?

lib/dataSources.js
anneeb: in case we want a return type that is not `ld+json`

package.json
anneeb: yep that's ok as long as it's not breaking any existing features in 8+

lib/dataSources.js
SpiralP: [`d2c4e12` (#276)](https://github.com/stardog-union/stardog.js/pull/276/commits/d2c4e12ba7bd3b9871d1076ce2ff4f6fd475cac7)

package.json
SpiralP: [`c0086a5` (#276)](https://github.com/stardog-union/stardog.js/pull/276/commits/c0086a5175d6b4b91c04059cec00e1216bb2b0c7)

anneeb (approved):
LGTM

alex-r-bigelow (approved):

PR #283, opened 2022-10-11
[VET-2546] **Breaking Change** Do not flatten request body nor unflatten response body for `db.options`

joshhk72:
[VET-2546](https://stardog.atlassian.net/browse/VET-2546)


SpiralP (approved):
insane to me that get() uses PUT and acts like getAll()/GET but is filtered to what you pass it

looks good to me!

PR #285, opened 2022-11-29
Adding addRole to the user code

iamstickfigure:
Fix typo in types file

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -831,7 +831,7 @@ declare namespace Stardog {
     export namespace user {
 
         interface User {
-            name: string;
+            username: string;
             password: string;
             superuser?: boolean;
         }
@@ -935,7 +935,7 @@ declare namespace Stardog {
          * @param {string} role the role to add to the user
          * @param {object} params additional parameters if needed
          */
-        function addRole(conn: Connection, username: string, role: string, params?: object): Promise<HTTP.Body>;
+        function assignRole(conn: Connection, username: string, role: string, params?: object): Promise<HTTP.Body>;
 
         /**
          * Gets a list of roles assigned to a user.
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -831,7 +831,7 @@ declare namespace Stardog {
     export namespace user {
 
         interface User {
-            username: string;
+            name: string;
             password: string;
             superuser?: boolean;
         }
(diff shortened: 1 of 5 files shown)


lib/index.d.ts
```diff
@@ -831,7 +831,7 @@ declare namespace Stardog {
     export namespace user {
 
         interface User {
-            username: string;
+            name: string;
```
anneeb, line 834: i know this is annoying, but can you revert this change? we have plans to address the inconsistency with this type in `VET-2366`

lib/index.d.ts
```diff
@@ -927,6 +927,16 @@ declare namespace Stardog {
...
 
+        /**
+         * Adds a role to a user.
+         * 
+         * @param {Connection} conn the Stardog server connection
+         * @param {string} username the username of the user
+         * @param {string} role the role to add to the user
+         * @param {object} params additional parameters if needed
+         */
+        function addRole(conn: Connection, username: string, role: string, params?: object): Promise<HTTP.Body>;
```
anneeb, line 938: what do you think about renaming this `assignRole` to match our current naming convention for `user.assignPermission` and `user.role.assignPermission`?

can you also update the docs by running the `docs` script?

lib/index.d.ts
joshhk72, line 834: This fix is correct, but we're tracking the "wrongness" in [this ticket](https://stardog.atlassian.net/browse/VET-2366). 

I'm wondering if we should also change this for `user.create` in this PR, or just record that this was fixed in the ticket.

iamstickfigure: I will make an effort to revert it. I recall this causing me a lot of pain yesterday, and I was shocked that the type had been wrong for such a long time. Lol. But I'll see if there's some workaround I can do to get typescript to look the other way.

So, I presume Stardog.js will eventually require this to be `username` rather than `name`? So I guess in my code, I'll set both `name` and `username` to the same value, so it future-proofs the upcoming fix. And I'll add a comment to fix the hack once VET-2366 is done

anneeb (approved):

SpiralP (approved):

anneeb (changes requested):
from running locally, we need to update the test to either use a role that exists in the test database by default, or seed the test database with a role that will be added as part of the test

anneeb (approved):

PR #289, opened 2022-12-20
[VET-2689] Add support for data source query options

alex-r-bigelow:
With the platform API change, we need to support sending JSON to the query endpoint

- [x] Support query + options object, in addition to strings
- [ ] Need to release (patch?) version

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -2,1506 +2,1979 @@
 
 /** stardog.js: The Stardog JS API*/
 
-import {
-    Headers,
-    Request,
-    RequestInit
-} from 'node-fetch';
+import { Headers, Request, RequestInit } from 'node-fetch';
 
 declare namespace Stardog {
-    namespace HTTP {
-        export type RdfMimeType = 'application/ld+json'
-            | 'text/turtle'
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -133,5 +133,5 @@
     "singleQuote": true,
     "trailingComma": "es5"
   },
-  "stardog-version": ">=8.0.0"
+  "stardog-version": ">=8.2.0"
```
anneeb, line 136: we should not change this-- this package is still compatible with Stardog 8.0.0+

lib/index.d.ts
```diff
@@ -1458,9 +1460,9 @@ declare namespace Stardog {
          *
          * @param {Connection} conn the Stardog server connection
          * @param {string} name the data source name
-         * @param {string} dataSourceQuery the data source query
+         * @param {string | DataSourceQuery} dataSourceQuery the data source query
```
anneeb, line 1463: it would be good to denote here somewhere that `DataSourceQuery` is only available on 8.2+

also, can you update the docs via `npm run docs` as well?

SpiralP (approved):

anneeb (approved):

PR #290, opened 2022-12-23
[VET-2824] Compatibility with node 18

SpiralP:
VET-2824

Node 18 comes with its own (experimental) fetch global which doesn't play nice with FormData, so fallback to using our fetch polyfill if running in node

### DIFF
--- a/lib/fetch.js
+++ b/lib/fetch.js
@@ -11,6 +11,14 @@ function hasAll(keys, obj) {
   return keys.every(key => key in obj);
 }
 
+function isNode() {
+  return typeof process !== 'undefined' &&
+    typeof process.release !== 'undefined' &&
+    typeof process.release.name !== 'undefined'
+    ? process.release.name === 'node'
+    : false;
+}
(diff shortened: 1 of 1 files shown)


lib/fetch.js
```diff
@@ -11,6 +11,14 @@ function hasAll(keys, obj) {
   return keys.every(key => key in obj);
 }
 
+function isNode() {
+  return typeof process !== 'undefined' &&
+    typeof process.release !== 'undefined' &&
+    typeof process.release.name !== 'undefined'
+    ? process.release.name === 'node'
+    : false;
+}
```
anneeb, line 20: is it worth only doing this specifically for node 18+?

anneeb (approved):

paulambro:
@jhartling reminder about this PR... looks relatively straightforward to review in short order :-)

jhartling (approved):
LGTM! I was able to load data in the stardog react tutorial with node 18.

PR #291, opened 2023-01-05
[VET-2831] Update Stardog.js to allow arbitrary metadata for stored queries

SpiralP:
VET-2831

* ~~now using the jsonld style because the json type is deprecated in stardog, and doesn't return or accept annotations~~
* removed the `params` args; what was this even used for?

in order to return the extra fields on stored queries you have to change `accept` on `stored.list` to use something besides json:
```js
stored.list(conn, { accept: 'application/ld+json' });
```
but note that this now returns jsonld and not simple json!

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -759,14 +759,14 @@ declare namespace Stardog {
              *
              * @param {Connection} conn the Stardog server connection
              * @param {StoredQueryOptions} storedQuery an object specifying the options to set on the updated query
-             * @param {boolean} useUpdateMethod whether to use Stardog's HTTP PUT method, added in version 6.2.0. Default: true
              * @param {object} options additional options to customize query
+             * @param {boolean} useUpdateMethod whether to use Stardog's HTTP PUT method, added in version 6.2.0. Default: true
              */
             function update(
                 conn: Connection,
                 storedQuery: StoredQueryOptions,
-                useUpdateMethod?: boolean = true,
-                options?: { accept?: string = 'application/json', contentType?: string = 'application/json' }
+                options?: { accept?: string = 'application/json', contentType?: string = 'application/json' },
+                useUpdateMethod?: boolean = true
             ): Promise<HTTP.Body>
 
             /**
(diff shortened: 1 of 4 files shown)


lib/query/stored.js
```diff
@@ -1,47 +1,78 @@
 const { fetch } = require('../fetch');
-const lodashPick = require('lodash/pick');
 const { httpBody } = require('../response-transforms');
 
+const convertToJsonLd = (conn, storedQuery) => {
```
SpiralP, line 4: i'm not really happy with this change; i think it's the best move for not adding a breaking change, but let me know if anyone has better ideas!

some ideas:
* add new methods where args are just passed through; ie accept, content-type, and storedQuery are just passed to fetch
* get rid of these json-style methods since they're deprecated (in stardog source code)

lib/query/stored.js
```diff
@@ -64,47 +65,37 @@ const renameStoredQuery = (conn, name, newName) => {
...
     return deleteResponse;
   });
 
 const updateStoredQuery = (
   conn,
   storedQuery,
-  params,
-  useUpdateMethod = true
+  useUpdateMethod = true,
+  options = {}
```
anneeb, line 81: this is a breaking change-- can we keep the args order consistent?

lib/index.d.ts
```diff
@@ -738,37 +738,44 @@ declare namespace Stardog {
              * Stores a query in Stardog, either on the system level or for a given database.
              *
              * @param {Connection} conn the Stardog server connection
-             * @param {StoredQueryOptions} config an object specifying the options to set on the new query
-             * @param {object} params additional parameters if needed
+             * @param {StoredQueryOptions} storedQuery an object specifying the options to set on the new query
+             * @param {object} options additional options to customize query
```
anneeb, line 742: not sure this description is quite accurate-- it's more like additional options to customize the request, or just request options now

SpiralP: [`d833110` (#291)](https://github.com/stardog-union/stardog.js/pull/291/commits/d833110c8d2444290c917019b3713e54eb4896f7) ?

anneeb (approved):

jhartling (approved):
LGTM!

PR #294, opened 2023-01-26
fix type for `stored.update`

anneeb:


SpiralP (approved):

README.md
```diff
@@ -1231,7 +1231,7 @@ Expects the following parameters:
 
 - conn ([`Connection`](#connection))
 
-- storedQuery ([`StoredQueryOptions`](#storedqueryoptions))
+- storedQuery ([`StoredQueryOptions | object`](#storedqueryoptions | object))
```
jhartling, line 1234: The hash is a link within the document which is broken now (other location is as well). It should be:
```
([`StoredQueryOptions`](#storedqueryoptions) | object)
```

anneeb: oof this is probably a bug elsewhere too then... let's file a separate ticket for fixing the docs generation

jhartling (approved):

PR #297, opened 2023-03-14
Cloud 1157 get server properties

iamstickfigure:


lib/server.js
```diff
@@ -25,7 +25,23 @@ const status = (conn, params = {}) => {
   ).then(httpBody);
 };
 
+const properties = (conn, names = []) => {
+  const headers = conn.headers();
+  headers.set('Accept', 'application/json');
+  const queryString = names.map(name => `name=${name}`).join('&');
```
SpiralP, line 31: should we use `encodeURIComponent` on `${name}` here?

SpiralP (approved):
1 suggestion above, but looks good to me; thanks!

lib/server.js
iamstickfigure: It's what the querystring package does, so it's probably good to be consistent. Good catch, thank you

anneeb:
released in https://github.com/stardog-union/stardog.js/releases/tag/v4.9.0 and https://github.com/stardog-union/stardog.js/releases/tag/v3.8.0

PR #302, opened 2023-08-28
[VET-3627] Add support for catalog endpoints

alex-r-bigelow:
VET-3627

### DIFF
--- a/lib/catalog.js
+++ b/lib/catalog.js
@@ -22,7 +22,7 @@ const reload = (conn, provider) => {
   }).then(httpBody);
 };
 
-const addCredentials = (conn, user, label) => {
+const addCredential = (conn, user, label) => {
   const headers = conn.headers();
   headers.set('Content-Type', 'application/json');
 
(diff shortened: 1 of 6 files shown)


lib/catalog.js
```diff
@@ -0,0 +1,68 @@
...
+  headers.set('Content-Type', 'application/json');
+
+  return fetch(conn.request('admin', 'catalog', 'reload'), {
+    method: 'POST',
+    body: JSON.stringify({ provider }),
+    headers,
+  }).then(httpBody);
+};
+
+const addCredentials = (conn, user, label) => {
```
alex-r-bigelow, line 25: Copied the spelling verbatim from [the platform code](https://github.com/stardog-union/stardog/blob/79a96e23dcfca24d145edeb96b6352249be128d4/catalog/protocols/http/client/main/src/com/complexible/stardog/catalog/protocols/http/client/HttpCatalogAdminConnectionImpl.java#L116) for now... but I kinda feel like this should be `addCredential` to be consistent with `removeCredential`? The endpoint itself is just `credentials`

SpiralP: i wouldn't mind if we named it `addCredential` here

SpiralP (approved):

anneeb (approved):

PR #304, opened 2024-01-09
[VET-4197] Add testConnection method for Knowledge Catalog

joshhk72:
VET-4197

No hurry to review/release since I can just use a local build 🤷

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -1415,6 +1415,14 @@ declare namespace Stardog {
          * @param {string} accessKey the accessKey that was assigned to the credential when it was added
          */
         function removeCredential(conn: Connection, accessKey: string): Promise<HTTP.Body>;
+
+        /**
+         * Tests a provider connection
+         *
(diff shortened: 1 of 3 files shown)


joshhk72:
Not quite sure I can test this since we'd need a valid connection set up.

mysticfalconvt (approved):

lib/index.d.ts
```diff
@@ -1415,6 +1415,14 @@ declare namespace Stardog {
          * @param {string} accessKey the accessKey that was assigned to the credential when it was added
          */
         function removeCredential(conn: Connection, accessKey: string): Promise<HTTP.Body>;
+
+        /**
+         * Tests a provider connection
```
SpiralP, line 1420:
```suggestion
         * Tests a provider connection using existing/new credentials.
```

`npm run docs` would be overwritten if not changed here as well

SpiralP (approved):
1 change above, but otherwise looks good to me, thanks!

PR #306, opened 2024-01-29
[VET-3766][VET-4297] Update addCredentials to support Atlan and Purview

hasibhassan:
VET-3766 VET-4297

### DIFF
--- a/lib/catalog.js
+++ b/lib/catalog.js
@@ -26,25 +26,10 @@ const addCredential = (conn, credentials, label) => {
   const headers = conn.headers();
   headers.set('Content-Type', 'application/json');
 
-  let body;
-  if (credentials.username) {
-    body = {
-      username: credentials.username,
-      password: credentials.password,
-      label,
-    };
-  } else if (credentials.token) {
-    body = {
-      token: credentials.token,
-      label,
-    };
-  } else {
-    body = {
-      clientId: credentials.clientId,
-      clientSecret: credentials.clientSecret,
-      label,
-    };
-  }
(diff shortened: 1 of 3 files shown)


lib/index.d.ts
```diff
@@ -1396,10 +1396,10 @@ declare namespace Stardog {
          * Adds a provider credential
          *
          * @param {Connection} conn the Stardog server connection
-         * @param {User} user the username and password to be stored
+         * @param {Credentials} credentials the credentials to be stored
```
joshhk72, line 1399: Need to also add `Credentials` type under `catalog` namespace (similar to how `DataSourceQuery` is under `dataSources` namespace.

lib/catalog.js
```diff
@@ -22,13 +22,12 @@ const reload = (conn, provider) => {
...
 
-const addCredential = (conn, user, label) => {
+const addCredential = (conn, credentials, label) => {
   const headers = conn.headers();
   headers.set('Content-Type', 'application/json');
 
   const body = {
-    username: user.username,
-    password: user.password,
+    ...credentials,
```
joshhk72, line 30: Works for me, but should we allow the user to set whatever on this? (I don't know if the way we do it normally is from inertia, or security, or etc.).

joshhk72 (approved):
LGTM, just some comments + small doc change

anneeb (approved):

anneeb:
released in https://github.com/stardog-union/stardog.js/releases/tag/v5.3.1

PR #307, opened 2024-04-19
[VET-4456] Add DS table endpoints

joshhk72:


test/dataSources.spec.js
```diff
@@ -142,6 +152,33 @@ describe('data_sources', () => {
         }));
   });
 
+  describe('getTables', () => {
```
joshhk72, line 155: Tests don't run (and never have for this file), but these should at least inform how it should be & what the responses look like.

anneeb (approved):

mysticfalconvt (approved):

PR #308, opened 2024-04-22
update tests for stardog v10

anneeb:


joshhk72 (approved):

SpiralP:
noting i did a light update to fetch-ponyfill (node-fetch dep) to fix the "socket hang up" bug that ci was seeing

this was happening because ci uses node 20 from the default circleci machine image (not docker), and ci needs to run as a machine because we run docker containers for stardog server, and i don't think you can use docker inside a docker circleci runtime

https://github.com/node-fetch/node-fetch/issues/1735

SpiralP (approved):

PR #314, opened 2025-08-12
[VET-5911] Improve VET app release scripts

SpiralP:
VET-5911


SpiralP:
https://github.com/stardog-union/stardog.js/pull/315

PR #315, opened 2025-08-22
[VET-5931] Upgrade Stardog.js to node 22

SpiralP:
VET-5931


.circleci/config.yml
SpiralP, line 1: will need to update github settings for pr blocking these 2 tasks, and remove required `build`

image

PR #318, opened 2025-11-11
[VET-6222] Investigate and fix issue with special characters in questions

SpiralP:
VET-6222

### DIFF
--- a/lib/catalog.js
+++ b/lib/catalog.js
@@ -10,7 +10,7 @@ const status = conn => {
 
 const reload = (conn, provider) => {
   const headers = conn.headers();
-  headers.set('Content-Type', 'application/json');
+  headers.set('Content-Type', 'application/json; charset=utf-8');
   return fetch(conn.request('admin', 'catalog', 'reload'), {
     method: 'POST',
     body: JSON.stringify({ provider }),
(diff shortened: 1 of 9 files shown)


lib/catalog.js
```diff
@@ -61,7 +61,7 @@ const runJob = (conn, jobname) => {
 
 const testConnection = (conn, connectionOptions) => {
   const headers = conn.headers();
-  headers.set('Content-Type', 'text/turtle');
```
jakehamilton, line 64: Should this change back to text/turtle when we eventually revert or is this call supposed to be json always?

jakehamilton (approved):
Thank you for the fix :)

lib/catalog.js
SpiralP: oh yea, fixed a bug here

below there is a:
```js
// connectionOptions is a JSON object made up of properties of metadata provider parameters and their values in Turtle syntax
body: JSON.stringify(connectionOptions),
```

and it's used like this in another app:
```js
  testConnection: createStardogAsyncThunk<{
    connectionOptions: Record<string, string>;
  }>('providers/testConnection', ({ connection, connectionOptions }) =>
    stardogCatalog.testConnection(connection, connectionOptions)
  ),
```
so it seems to be a json object of `Record<string, string>` (i think turtle was a mistake)

SpiralP: i should also mention i did test this, and both application/json and text/turtle give the same result with the same input

reading stardog code, it seems to do `JSON.parse(...).getAsJsonObject()` first as well

anneeb (approved):

PR #319, opened 2025-11-21
[VET-6222] Investigate and fix issue with special characters in questions

SpiralP:
VET-6222 VET-<span>6163</span>

`dataSources.query` can't accept `application/json; charset=utf-8`, but does work with `application/json`:

```
curl ...
        -H 'content-type: application/json; charset=utf8' \
        --data-raw '{"query":"select * from PURCHASE_ORDER where PO_DATE > \'2020-04-01\' LIMIT 25","options":{"parser.sql.quoting":"NATIVE","sql.functions":"","percent.encode":true}}'
Data Source query failed: 'pilothouse-test-new-ds-154452d5-11eb-4030-858b-fdcadc384cd4': Invalid SQL syntax found. Encountered "{" at line 1, column 1.
Was expecting one of:
```


test/dataSources.spec.js
```diff
@@ -1,58 +1,54 @@
...
 const { ConnectionFactory } = require('./setup-database');
+const snapshots = require('./__snapshots__/dataSources.spec.js.snap');
 
-const dataSourceMetadata = fs.readFileSync(
-  path.resolve(`${__dirname}/fixtures/data_source_metadata.ttl`),
-  'utf8'
-);
+/*
+Run the following to set up the MySQL database used in these tests:
+echo "drop database if exists stardogjs_test; create database stardogjs_test; use stardogjs_test; create table test_table (id INT PRIMARY KEY);" | mysql -h host.docker.internal -uroot -pmy-secret
```
SpiralP, line 9: noting i got all these here to pass after running this (and removing the `.only` below)

anneeb (approved):

mysticfalconvt (approved):

PR #321, opened 2026-07-30
[VET-7246] Add dataSources.test method

SpiralP:
[VET-7246](https://stardog.atlassian.net/browse/VET-7246) (parent: [VET-7013](https://stardog.atlassian.net/browse/VET-7013))

## Summary

Adds [`dataSources.test(conn, name)`](https://github.com/stardog-union/stardog.js/blob/82ecd1a4013784a2fa5c39c7f59162282f547f2b/lib/dataSources.js#L99-L109), a `POST` to `admin/data_sources/{name}/test_data_source`, so Studio's new "Connection test" card can check whether a saved data source is actually reachable instead of hand-rolling the fetch and the response transform. Also declares `Connection#request` in the type definitions — a method that has always existed and that every method in this library calls, but which TS consumers had to cast to reach.

## What changed

- **`dataSources.test`** — modeled on the neighbouring [`online`](https://github.com/stardog-union/stardog.js/blob/82ecd1a4013784a2fa5c39c7f59162282f547f2b/lib/dataSources.js#L82-L89): same verb, same `Accept: application/json`, same `httpBody` transform. A reachable source answers `204` (so [`httpBody` nulls the body](https://github.com/stardog-union/stardog.js/blob/82ecd1a4013784a2fa5c39c7f59162282f547f2b/lib/response-transforms.js#L22-L24)); an unreachable one is expected to answer non-2xx with the server's error text in `res.body`, which is what the Studio failure block renders (unverified — see below).
- **Why `httpBody` and not the raw `Response`** — `httpBody` picks the response fields onto a plain object. Consumers that check own enumerable keys to tell a Stardog response apart from something else see this method like any other. Returning a bare fetch `Response` would leave `ok` on the prototype and make failures read as successes.
- **Type definitions** — the `test` declaration, plus [`Connection#request(...resource): string | Request`](https://github.com/stardog-union/stardog.js/blob/82ecd1a4013784a2fa5c39c7f59162282f547f2b/lib/index.d.ts#L120). `uri` was declared and `request` never was, even though the declarations already describe the `meta.createRequest` hook that [only `request` consumes](https://github.com/stardog-union/stardog.js/blob/82ecd1a4013784a2fa5c39c7f59162282f547f2b/lib/connection.js#L49-L66).
- **Test** — a `describe('test')` block asserting `ok`/`204` for a reachable source. It sits behind the spec's pre-existing [`describe.only('listInfo')`](https://github.com/stardog-union/stardog.js/blob/82ecd1a4013784a2fa5c39c7f59162282f547f2b/test/dataSources.spec.js#L62-L64), so like the rest of this file it does not run in CI — CI brings up a Stardog container but no MySQL service for the fixture.

## Acceptance criteria (VET-7246)

VET-7246 is the "Update stardog.js" sub-task and carries no criteria of its own; the parent's criteria are all Studio UI. The two that bear on this repo:

- [x] No new backend contract — this rides the existing `data-source test` endpoint (`stardog-admin data-source test <name>`).
- [x] A caller can tell pass from fail and, on fail, get the server-provided error message to display.
- [ ] Connection test card, in-flight/pass/fail states, unsaved-changes tooltip, copy button, Stardog 8+ gate — all Studio-side, separate PR.

## How to verify

**Setup:** a local Stardog (the repo `Dockerfile`) plus the MySQL fixture named in the [comment at the top of the spec](https://github.com/stardog-union/stardog.js/blob/82ecd1a4013784a2fa5c39c7f59162282f547f2b/test/dataSources.spec.js#L7-L18):

```
echo "drop database if exists stardogjs_test; create database stardogjs_test; use stardogjs_test; create table test_table (id INT PRIMARY KEY);" | mysql -h host.docker.internal -uroot -pmy-secret
```

- Move the `.only` from `listInfo` onto `describe('test')` (or drop it) and run `yarn run test:unit --testPathPattern dataSources` — the happy path should be `ok: true`, `status: 204`. Left as-is, the new block is silently skipped.
- **Failure path is the interesting one and is not covered by a test.** Point the data source at a dead JDBC URL (`dataSources.update` with a bad host or password), call `test`, and confirm the rejection surfaces as a resolved promise with `ok: false` and a human-readable message in `res.body` — that string is what Studio puts in the copyable error block. Worth eyeballing the shape (JSON `{message}` vs. plain text) since the Studio side has to render it.
- Sanity-check that the endpoint path is right: `test_data_source`, not `test`. The siblings in this file are `available` / `online` / `share`, so the `_data_source` suffix is the odd one out — it matches the server route, but it is the one thing here that would fail silently as a 404 rather than loudly.
- `Connection#request` is a types-only change; `yarn run build` is enough to confirm it compiles, and no runtime behavior moves.

## Notes for reviewers

- **Long-running tests have no lever here.** The parent's grooming notes flag sources (Databricks was the example) where the server-side test never comes back in reasonable time. This method inherits whatever `fetch` does and adds no timeout or cancellation — a caller that needs one has to go through `meta.createRequest`. Fine for this PR if we agree that belongs to the consumer, but worth a conscious call.
- **No version gating in the library**, consistent with the rest of it. The Stardog 8+ gate from the parent ticket lives in Studio.
- README API docs are generated from `lib/index.d.ts` by `yarn run docs`, and the changelog by `scripts/changelog.js` — both run during `yarn version`, so neither is hand-edited here.

[VET-7246]: https://stardog.atlassian.net/browse/VET-7246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[VET-7013]: https://stardog.atlassian.net/browse/VET-7013?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,6 +31,15 @@ workflows:
             - run:
                 name: Wait for Stardog server to be ready
                 command: yarn run wait-for-server
+            # `when: always` so the logs are still collected when the wait
+            # above fails, which is the only time they matter.
+            - run:
+                name: Show Stardog container logs
+                when: always
+                command: |
+                  docker ps -a
+                  docker logs stargoose || true
+                  docker cp stargoose:/var/opt/stardog/stardog.log /tmp/stardog.log && cat /tmp/stardog.log || true
       - node/run:
           name: Build with Node.js 22
           executor: machine
--- a/README.md
+++ b/README.md
@@ -227,6 +227,12 @@ Returns [`void`](#void)
 ### <a name="headers">Connection.headers()</a>
 
 Returns [`Headers`](#headers)
+### <a name="request">Connection.request(resource)</a>
+
+Takes the following params:
+- resource (`string[]`)
+
+Returns `string | Request`
 ### <a name="uri">Connection.uri(resource)</a>
 
 Takes the following params:
@@ -2417,6 +2423,18 @@ Returns [`Promise<HTTP.Body>`](#body)
 #### <a name="available">`dataSources.available(conn, name)`</a>
 
 Determine if the named data source is available
+
+Expects the following parameters:
+
+- conn ([`Connection`](#connection))
+
+- name (`string`)
+
+Returns [`Promise<HTTP.Body>`](#body)
+
+#### <a name="test">`dataSources.test(conn, name)`</a>
+
+Test the connection for the named data source
 
 Expects the following parameters:
 
--- a/lib/dataSources.js
+++ b/lib/dataSources.js
@@ -96,6 +96,18 @@ const available = (conn, name) => {
   }).then(httpBody);
 };
 
+const test = (conn, name) => {
+  const headers = conn.headers();
+  headers.set('Accept', 'application/json');
+  return fetch(
+    conn.request('admin', 'data_sources', name, 'test_data_source'),
+    {
+      method: 'POST',
+      headers,
+    }
+  ).then(httpBody);
+};
+
 const options = (conn, name) => {
   const headers = conn.headers();
   return fetch(conn.request('admin', 'data_sources', name, 'options'), {
@@ -240,6 +252,7 @@ module.exports = {
   remove,
   share,
   suggestions,
+  test,
   typeDescription,
   update,
   updateMetadata,
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -2016,6 +2016,14 @@ declare namespace Stardog {
     function available(conn: Connection, name: string): Promise<HTTP.Body>;
 
     /**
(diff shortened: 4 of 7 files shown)


.circleci/config.yml
SpiralP, line 1: ci tests failing because of "Your Stardog license has expired"

anneeb (approved):

mysticfalconvt (approved):

mysticfalconvt (approved):

